PR #56, opened 2020-02-18
Report form styling

liywjl:
Report form UI dialog needed some styling, so did the report button

## Links to tickets
https://trello.com/c/VuaxwXy0/1174-style-report-link
https://trello.com/c/gJZtH0E9/1175-show-report-form


src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -97,97 +97,144 @@ export const Form: React.FC<{
...
+    if (!e.shiftKey && document.activeElement === lastElement) {
+      firstElement.focus();
+      return e.preventDefault();
+    }
+
+    if (e.shiftKey && document.activeElement === firstElement) {
+      lastElement.focus();
+      e.preventDefault();
+    }
+  };
```
oliverlloyd: This is pretty narly. We could help future developers with comments explaining what the reasons and intentions are here with maybe some links to the supporting articles?

gtrufitt:
What are your thoughts on making this an uncontrolled form? It feels like there's a lot of complexity around the focus and keyboard navigation which is caused so that we can track state, rather than just handle form validation on submit. 

Let me know if I'm missing something!

I also wouldn't be opposed to bringing something well documented like https://react-hook-form.com/ (https://bundlephobia.com/result?p=react-hook-form@4.9.6) to handle the forms and validation within discussion rendering rather than writing ourselves? What do you reckon?

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -63,100 +94,154 @@ export const Form: React.FC<{
...
+
+  const keyListenersMap = new Map([
+    [27, toggleSetShowForm],
```
gtrufitt: The `esc` key binding here is a nice touch! Maybe you could set focus on the report button so the user is back to where they might expect when the form is closed (either by esc or close button)

liywjl:
> What are your thoughts on making this an uncontrolled form? It feels like there's a lot of complexity around the focus and keyboard navigation which is caused so that we can track state, rather than just handle form validation on submit.
> 
> Let me know if I'm missing something! I did check the branch out to get a clearer understanding and it seems that if I try and type in the form it sends my focus elsewhere - I guess that's an example of why we might want to not try and handle all that behaviour!
> 
> I also wouldn't be opposed to bringing something well documented like https://react-hook-form.com/ (https://bundlephobia.com/result?p=react-hook-form@4.9.6) to handle the forms and validation within discussion rendering rather than writing ourselves? What do you reckon?

I think bringing something like https://react-hook-form.com/  would be a great idea. I will look at simplifying the logic here. I was following an online example on accessibility, but feel that there is quite a couple of areas were that could be simplified.

gtrufitt:
> I was following an online example on accessibility, but feel that there is quite a couple of areas were that could be simplified.

Ah yeah, actually I was misunderstanding something, partly because I was looking at the form in isolation of storybook :)

I guess the thing I wasn't quite understanding was why the focus stuff was required - I realised it was to trap the focus state within the modal, is that correct? In which case, that makes sense, the focus stuff isn't form related as such but modal related (and not react related but modal related).

liywjl:
> ct? In which case, that makes sense, the focus stuff isn't form related as such but modal related (

Yes this is all about keeping the focus in the modal. But I see some areas were this logic can be greatly simplified. I think the hooks form lib would also allow us to add error handling without too much complexity

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -1,31 +1,46 @@
...
+  right: 0;
+  display: flex;
+  flex-direction: column;
   padding: ${space[3]}px;
   background-color: white;
   ${textSans.xsmall()};
 `;
 
+const labelStyles = (pillar: Pillar) => css`
+  color: ${palette[pillar][400]};
```
oliverlloyd: What's 400? Is this something new? I thought pillar colours were named.

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -44,175 +59,197 @@ const inputWrapper = css`
...
-    </div>
+  pillar: Pillar;
+}> = ({ commentId, toggleSetShowForm, pillar }) => {
+  // We want to listen to keydown events for accessibility
+  useEffect(() => {
+    const keyListenersMap: { [key: number]: any } = {
+      27: toggleSetShowForm,
+      9: handleTabKey
+    };
+    const keyListener = (e: KeyboardEvent) => {
```
oliverlloyd: Do we need this abstraction if we're only caring about two keys?

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -44,175 +59,197 @@ const inputWrapper = css`
...
-}> = ({ commentId }: Props) => {
-  const [showForm, setShowForm] = useState(false);
-  const toggleSetShowForm = () => setShowForm(!showForm);
+  // This function only gets called when a tab key has been pressed
+  // we use `e.shiftKey` internally to determin the direction of the highlighting
+  const handleTabKey = (e: KeyboardEvent) => {
+    if (!e.shiftKey && document.activeElement === lastElement) {
+      firstElement && firstElement.focus();
+      e.preventDefault();
+    }
```
oliverlloyd: I completely got this on first read 👍

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -44,175 +59,197 @@ const inputWrapper = css`
...
 
-  const [errors, setErrors] = useState({});
+    if (e.shiftKey && document.activeElement === firstElement) {
+      lastElement && lastElement.focus();
```
oliverlloyd:
```suggestion
      lastElement && lastElement.focus(); // The shift key is down so loop focus back to the last item
```

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -44,175 +59,197 @@ const inputWrapper = css`
...
-  const [formState, setFormState] = useState({
-    categoryId: 0,
-    reason: "",
-    email: ""
+  const { register, handleSubmit, errors } = useForm({
+    defaultValues: {
+      categoryId: 0,
+      reason: "",
+      email: ""
+    }
```
oliverlloyd: Are these used? I guess they could be part of the form lib but if not, maybe set default when destructuring below?

src/components/Comment/Comment.tsx
```diff
@@ -14,24 +14,23 @@ type Props = {
...
+  margin-right: ${space[2]}px;
+  color: ${palette[pillar][400]};
+  border: 0;
+`;
 
-  *:last-child {
-    float: right;
-    font-weight: normal;
-  }
+const controleWrapper = css`
```
oliverlloyd: Nitpick: it's 'control', not 'controle'. But maybe 'spaceBetween' is more descriptive?

src/components/Comment/Comment.tsx
```diff
@@ -89,6 +88,7 @@ export const avatar = (avatarSize: number): string => css`
 `;
 
 export const Comment = ({ comment, pillar }: Props) => {
+  const commentControlsButtonStyles = commentControlsButton(pillar);
```
oliverlloyd: I'd prefer removing this abstraction and just inlining the function call. It will make it easier to trace the code with one less step to follow

oliverlloyd:
A few points (based on our chat irl)

1. The parent needs to control which form is open to prevent multiple modals showing at the same time
2. We want to set focus back to the report link that opened the modal when the modal is closed. Maybe we could use an `onClose()` callback to have the parent manage this?
3. `react-form` is nice, but 8kb is a bit hefty...

gtrufitt:
> https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset

I'm not sure whether it'll be able to tree shake smaller. I just wonder whether when all is said and done with forms functionality we'll not have written 8kb of code anyway 🤷‍♂

liywjl:
> > react-form is nice, but 8kb is a bit hefty...
> 
> I'm not sure whether it'll be able to tree shake smaller. I just wonder whether when all is said and done with forms functionality we'll not have written 8kb of code anyway 🤷‍♂

I think @oliverlloyd is right here. We only have somewhat complex form logic in the report form. It probably doesn't merit adding and extra lib. I think we can get away with maybe removing it.

I noticed we are able to open multiple models here, so will fix

gtrufitt:
> I think we can get away with maybe removing it.

Cool. Not like we have other forms on the site that would reuse it either!

src/components/AbuseReportForm/AbuseReportForm.tsx
liywjl: it is used elsewhere, and would make the design look constant. I think it is safe to keep. We can reopen this later when we run through final design

src/components/Comment/Comment.tsx
liywjl: this is a minor optimisation, but defining it once here we avoid having 3 different css classes that do the same thing.

oliverlloyd: Ah, I see what you mean. I suspect though that each call to the function will return the same string - emotion (as far as I remember without trawling the interwebs) hashes the css to get the classname so if the css is the same each time, so is the classname.

If not, then we probably want to have a conversation about the best way to make this optimisation and if we think the extra cognitive load of the abstraction is worth it - there's lots of places in the codebase where we could do this but currently don't

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -44,175 +58,243 @@ const inputWrapper = css`
...
+        categoryId: "You must select a category before submitting"
+      });
+    }
+    const response = await reportAbuse({
+      categoryId,
+      reason,
+      email,
+      commentId
+    });
+    if (response.status !== 200) {
```
oliverlloyd: I think you can use `if (!response.ok) ...` here to get better coverage

PR #58, opened 2020-02-19
Filter with new styles

liywjl:
### What?

We have new designs for how the filter should look on DCR.

- Mobile
Screenshot 2020-02-20 at 10 51 33

- Desktop
Screenshot 2020-02-20 at 10 51 14

### Why?

Bec they look nicer and should be easier for users of small devices to use 

### Trello link
https://trello.com/c/G4XtiClt/1172-style-filters

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -34,6 +34,7 @@ export const App = ({ shortUrl, user }: Props) => {
     threads: "unthreaded",
     page: 1
   });
+  const [commentCount, useCommentCount] = useState<number>(0);
   const [loading, setLoading] = useState<boolean>(true);
   const [pages, setPages] = useState<number>(0);
 
@@ -84,6 +85,8 @@ export const App = ({ shortUrl, user }: Props) => {
       setComments(json.discussion.comments);
       setPages(json.pages);
     });
+
+    // TODO: get getCommentCount to work
   }, [filters, shortUrl]);
 
   console.log("comments", comments);
@@ -94,7 +97,12 @@ export const App = ({ shortUrl, user }: Props) => {
         <CommentForm shortUrl={shortUrl} onAdd={commentAdded} user={user} />
       )}
       <TopPicks shortUrl={shortUrl} />
-      <Filters filters={filters} setFilters={setFilters} pages={pages} />
+      <Filters
+        filters={filters}
+        setFilters={setFilters}
+        pages={pages}
+        commentCount={commentCount}
+      />
       {loading ? (
         <p>TODO loading component goes here...</p>
       ) : (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,9 +2,7 @@ import React, { useState, useEffect } from "react";
 import { css } from "emotion";
 
 import { CommentType, FilterOptions, UserProfile } from "./types";
-
 import { getDiscussion, getCommentCount } from "./lib/api";
-
 import { CommentList } from "./components/CommentList/CommentList";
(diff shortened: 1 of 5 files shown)


gtrufitt:
Could we have screenshots? :)

src/components/Filters/Filters.tsx
```diff
@@ -71,22 +64,33 @@ const filterContainer = css`
...
+const paginationWrapper = css`
   display: flex;
   flex-direction: row;
-  @media screen and (max-width: 740px) {
+  justify-content: space-between;
+  width: 100%;
+  margin-top: 15px;
+  padding-top: 10px;
+  border-top: 1px solid #dcdcdc;
+  @media screen and (max-width: 480px) {
```
oliverlloyd: We should get breakpoints and colours from source foundation

src/components/Filters/Filters.tsx
```diff
@@ -170,16 +174,22 @@ export const Filters = ({ filters, setFilters, pages }: Props) => (
...
+            setFilters({
+              ...filters,
+              page: page
+            });
+          }}
+        />
+      </div>
+      <div className={paginationText}>
+        {/* TODO: pull text dynamically */}
+        Displaying 100 of 500 comments
```
oliverlloyd:
```suggestion
        TODO: Displaying 100 of 500 comments
```

oliverlloyd: Arguably, we shouldn't commit this if it's not working but If we do commit hardcoded text then it should be explicit that it is not real

gtrufitt:
oh they look looooovely.

src/components/Filters/Filters.tsx
```diff
@@ -169,17 +180,27 @@ export const Filters = ({ filters, setFilters, pages }: Props) => (
...
+                page: page
+              });
+            }}
+          />
+        </div>
+        <div className={paginationText}>
+          {commentCount &&
+            `Displaying ${filters.pageSize *
+              filters.page} of ${commentCount} comments`}
+        </div>
```
oliverlloyd: There are no designs for what appears at the footer of the list but assuming we show pagination in the same way, with this text, then we might want to think about pulling this code out.

Might also need to guard against multiply by zero errors and other edge cases

liywjl: technically the `filters.page < 2` should avoid any multiply by zero errors

src/App.tsx
```diff
@@ -84,6 +85,8 @@ export const App = ({ shortUrl, user }: Props) => {
       setComments(json.discussion.comments);
       setPages(json.pages);
     });
+
+    // TODO: get getCommentCount to work
   }, [filters, shortUrl]);
 
   console.log("comments", comments);
```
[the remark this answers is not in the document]
liywjl: I removed it in another PR https://github.com/guardian/discussion-rendering/pull/56/files#diff-46ef2374da6211486f3a408db68b2bcbL89

src/App.tsx
```diff
@@ -86,6 +87,14 @@ export const App = ({ shortUrl, user }: Props) => {
     });
   }, [filters, shortUrl]);
 
+  useEffect(() => {
+    setLoading(true);
```
liywjl: technically this will cause a race condition with: https://github.com/guardian/discussion-rendering/pull/58/files#diff-46ef2374da6211486f3a408db68b2bcbR82

but this is beyond the scope of this PR

PR #59, opened 2020-02-24
Timestamp as a link

oliverlloyd:
Here we make Timestamp a link with a required `linkTo` prop

We then add `Timestamp` to `Comment` alongside the reader's username

Screenshot 2020-02-24 at 17 42 43

### Trello
https://trello.com/c/ur3ARTeI/1183-date-ago-is-a-link

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -8,6 +8,7 @@ import { Pillar, CommentType } from "../../types";
 import { GuardianStaff, GuardianPick } from "../Badges/Badges";
 import { RecommendationCount } from "../RecommendationCount/RecommendationCount";
 import { AbuseReportForm } from "../AbuseReportForm/AbuseReportForm";
+import { Timestamp } from "../Timestamp/Timestamp";
 
 type Props = {
   comment: CommentType;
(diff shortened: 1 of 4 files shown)


src/components/Timestamp/Timestamp.tsx
```diff
@@ -7,30 +7,30 @@ import { DateFromISOStringC } from "io-ts-types/lib/DateFromISOString";
...
+  linkTo: string;
 };
 
-const containerStyles = css`
-  display: flex;
-  flex-direction: row;
+const linkStyles = css`
+  color: ${palette.neutral[46]};
+  text-decoration: none;
+  :hover {
```
gtrufitt:
```suggestion
  :hover,
  :focus {
```

src/components/Timestamp/Timestamp.tsx
```diff
@@ -7,30 +7,30 @@ import { DateFromISOStringC } from "io-ts-types/lib/DateFromISOString";
...
-}: {
-  isoDateTime: DateFromISOStringC;
-}) => {
+export const Timestamp = ({ isoDateTime, linkTo }: Props) => {
   return (
-    <div className={containerStyles}>
-      <div className={countStyles}>{dateFormatter(isoDateTime)}</div>
-    </div>
+    <a href={linkTo} className={linkStyles}>
+      <time dateTime={isoDateTime.toString()} className={timeStyles}>
```
gtrufitt: 🆒

src/components/Comment/Comment.tsx
```diff
@@ -82,12 +83,41 @@ const iconWrapper = css`
...
+  justify-content: center;
+`;
+
 export const avatar = (avatarSize: number): string => css`
   border-radius: ${avatarSize + 10}px;
   width: ${avatarSize}px;
   height: ${avatarSize}px;
 `;
 
+const Column = ({ children }: { children: JSX.Element | JSX.Element[] }) => (
```
gtrufitt: Here we go again :) Probably can refactor other areas to use this as a util in another PR.

src/components/Comment/Comment.tsx
```diff
@@ -100,21 +130,19 @@ export const Comment = ({ comment, pillar }: Props) => {
...
-              <div className={commentProfileName(pillar)}>
-                {comment.userProfile.displayName}
-              </div>
-              <div
-                className={css`
-                  display: flex;
-                  flex-direction: row;
-                `}
-              >
+            <Column>
```
gtrufitt: Defo easier to reason with

src/components/Comment/Comment.tsx
oliverlloyd: Ha, yeah, I had that thought. I added it here to make the html more readable but it's really only useful where you have identical css, which isn't that often.

Aside: It would be more useful if we went with the `const PinkColumn = styled(Column)'color: pink; '` pattern so we can express what I want via composition.

PR #63, opened 2020-02-26
Reply to comment form

liywjl:
### What?
Add a comment reply form

### Why?
To be able to support reply forms

### How?
We decided after trying a couple of changes to keep the API call logic within the CommentForm component. This avoids unnecessary prop drilling.

Also as Comment component now is called recursively we make some assumptions if the comment is nested or not based on the existence of `parentDisplayReplyForm`

Similarly we also now determine if we call is `reply` or `comment` based on the existence of `replyComment` in CommentForm.


src/components/CommentList/CommentList.tsx
```diff
@@ -1,36 +1,173 @@
...
 
 type Props = {
-  comments?: CommentType[];
+  comment: CommentType;
+  pillar: Pillar;
+  shortUrl: string;
+  user?: UserProfile;
+  onAddComment: (commentId: number, body: string, user: UserProfile) => void;
   threads: ThreadsType;
+  replyComment?: CommentType;
```
oliverlloyd: So my personal preference is to not pass around objects. I would prefer a simple api where we pass just the id and, in the case where we need more information, we read it from either `comment` or `comment.responses` - but that's me 🤷‍♂

PR #64, opened 2020-02-26
Persist filters to local storage

oliverlloyd:
### What
This change ensures that the reader's filter settings are persisted to localStorage

### Why
So that next time they visit we will remember the settings they had before

### Trello
https://trello.com/c/1bes5LGl/1215-remember-filters

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,14 +24,63 @@ const footerStyles = css`
   justify-content: flex-end;
 `;
 
+const DEFAULT_FILTERS: FilterOptions = {
+  orderBy: "newest",
+  pageSize: 25,
+  threads: "collapsed",
+  page: 1
+};
+
+const rememberFilters = (filtersToRemember: FilterOptions) => {
+  try {
+    localStorage.setItem(
+      "gu.prefs.discussioni.threading",
+      JSON.stringify({ value: filtersToRemember.threads })
+    );
+    localStorage.setItem(
+      "gu.prefs.discussioni.pagesize",
+      JSON.stringify({ value: filtersToRemember.pageSize })
+    );
+    localStorage.setItem(
+      "gu.prefs.discussioni.order",
+      JSON.stringify({ value: filtersToRemember.orderBy })
+    );
+  } catch (error) {
+    // Sometimes it's not possible to access localStorage, we accept this and don't want to
+    // capture these errors
+  }
+};
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -80,19 +129,24 @@ export const App = ({ shortUrl, user }: Props) => {
     setLoading(true);
     getDiscussion(shortUrl, filters).then(json => {
       setLoading(false);
-      setComments(json.discussion.comments);
-      setPages(json.pages);
+      setComments(json?.discussion?.comments);
```
liywjl: I am going to assume if this is here that this feature works? I know my IDE always complains about it so I am never sure

oliverlloyd: It fixed the bug I got when I tried to read the value when it was undefined so it works in that sense!

oliverlloyd:
> would be awesome to do something similar to `rememberFilters` in dotcom-rendering to we dnt get local storage errors

Yes, totally agree. It adds a fair bit of extra code to handle the edge cases but this seems like a pattern worth having. I expect we'll be abstracting this out at some point but I'm happy to hold off from this until we have more use cases and understand how we want to use LS better.

src/App.tsx
oliverlloyd: That said ^, I'll check we don't have an issue when integrating with DCR

src/App.tsx
```diff
@@ -24,14 +24,63 @@ const footerStyles = css`
...
+const DEFAULT_FILTERS: FilterOptions = {
+  orderBy: "newest",
+  pageSize: 25,
+  threads: "collapsed",
+  page: 1
+};
+
+const rememberFilters = (filtersToRemember: FilterOptions) => {
+  try {
+    localStorage.setItem(
```
gtrufitt: Sorry, know this is merged but @oliverlloyd @liywjl I had a thought last night that we should try and think about how we create an abstraction layer over the *browser* specific things - like localstorage, so that they can hook into other technologies - apps for example might want to store these in the native layer so we should define a contract where we can plug in localstorage but others can plugin whatever mechanism works for them. No need to change this just yet but something to think about when building!

oliverlloyd: Yep, I agree (I made a comment about this ^). Probably worth reviewing FE to see what learnings are available and possibly having a look at libs like localForage to see how they handle edge cases.

gtrufitt: oh yeh. cool!

PR #67, opened 2020-02-27
Reply preview comment

liywjl:
### What?

Add preview comments to the top of reply form

Screenshot 2020-02-28 at 10 09 07

### Trello

https://trello.com/c/sTLY8sP7/1222-reply-header-preview

### DIFF
--- a/src/components/CommentReplyPreview/CommentReplyPreview.stories.tsx
+++ b/src/components/CommentReplyPreview/CommentReplyPreview.stories.tsx
@@ -1,0 +1,43 @@
+import React from "react";
+
+import { CommentReplyPreview } from "./CommentReplyPreview";
+import { CommentType } from "../../types";
+
+export default { title: "CommentReplyPreview" };
+
+const replyComment: CommentType = {
+  id: 25487686,
+  body:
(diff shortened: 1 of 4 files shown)


src/components/CommentReplyPreview/CommentReplyPreview.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+    commentCount: 2,
+    staffCommenterCount: 1,
+    editorsPickCount: 0,
+    blockedCount: 0,
+    responseCount: 1
+  }
+};
+
+export const Default = () => (
+  <CommentReplyPreview replyComment={replyComment} pillar={"sport"} />
```
oliverlloyd: These props look wrong, I think that should be `commentBeingRepliedTo`, not `replyComment`. Stuff like will be caught by linting in future once I complete a PR I'm working on!

src/components/CommentReplyPreview/CommentReplyPreview.tsx
```diff
@@ -0,0 +1,127 @@
...
+        </div>
+        <div className={replyDisplayNameStyles}>
+          {commentBeingRepliedTo.userProfile.displayName}
+        </div>
+        <button
+          className={cx(hideCommentButtonStyles, commentControlsButtonStyles)}
+          onClick={() => setDisplayReplyComment(!displayReplyComment)}
+        >
+          {displayReplyComment ? "Hide Comment" : "Show comment"}
+        </button>
```
oliverlloyd: On frontend they're using faux links here. I'm guessing the reasoning for that is to prevent focus landing on these controls because we don't want this part of the comment flow to be accessible.

PR #71, opened 2020-02-28
Staff pick

liywjl:
### What?

When a user is of type staff, they should be able to see the `pick` button and are able to pick comments

Screenshot 2020-02-28 at 16 04 14

### Trello
https://trello.com/c/AauPx6Q2/1178-staff-can-pick-comments

### DIFF
--- a/src/components/Comment/Comment.stories.tsx
+++ b/src/components/Comment/Comment.stories.tsx
@@ -71,7 +71,7 @@ export const ReplyComment = () => (
 );
 Default.story = { name: "default" };
 
-export const StaffUserPicked = () => (
+export const StaffUserUnpicked = () => (
   <Comment
     comment={commentData}
     pillar={"sport"}
@@ -80,9 +80,9 @@ export const StaffUserPicked = () => (
     isReply={false}
   />
 );
-StaffUserPicked.story = { name: "Staff user picked" };
+StaffUserUnpicked.story = { name: "Staff user unpicked" };
 
-export const StaffUserUnpicked = () => (
+export const StaffUserPicked = () => (
   <Comment
     comment={{ ...commentData, isHighlighted: true }}
     pillar={"sport"}
@@ -91,4 +91,4 @@ export const StaffUserUnpicked = () => (
     isReply={false}
   />
 );
-StaffUserUnpicked.story = { name: "Staff user unpicked" };
+StaffUserPicked.story = { name: "Staff user picked" };
--- a/src/components/Comment/Comment.stories.tsx
+++ b/src/components/Comment/Comment.stories.tsx
@@ -36,11 +36,36 @@ const commentData: CommentType = {
   }
 };
 
-export const defaultStory = () => (
+const staffUser = {
+  userId: "abc123",
+  displayName: "Jane Smith",
+  webUrl: "",
+  apiUrl: "",
+  avatar: "",
+  secureAvatarUrl: "",
+  badge: [{ name: "Staff" }],
+  privateFields: {
+    canPostComment: true,
+    isPremoderated: false,
+    hasCommented: true
+  }
+};
+
+export const Default = () => (
   <Comment
     comment={commentData}
     pillar={"sport"}
     setCommentBeingRepliedTo={() => {}}
   />
 );
-defaultStory.story = { name: "default" };
+Default.story = { name: "default" };
+
+export const StaffUser = () => (
+  <Comment
+    comment={commentData}
+    pillar={"sport"}
+    setCommentBeingRepliedTo={() => {}}
+    user={staffUser}
+  />
+);
+StaffUser.story = { name: "StaffUser" };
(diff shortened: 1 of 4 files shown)


src/components/Comment/Comment.tsx
```diff
@@ -172,14 +200,24 @@ export const Comment = ({
...
-              <button className={commentControlsButtonStyles}>Pick</button>
+              {/* Only staff can pick, and they cannot pick thier own comment */}
+              {user &&
+                user.badge.some(e => e.name === "Staff") &&
+                user.userId !== scopedComment.userProfile.userId && (
+                  <button
+                    onClick={pick}
+                    className={commentControlsButtonStyles}
+                  >
+                    {scopedComment.isHighlighted ? "Unpick" : "Pick"}
```
oliverlloyd: If this PR isn't adding the ability to unpick a pick we shouldn't add the UI for that action. Or if we do it should be made clear that the logic is pending to help future developers / prevent it being missed. Maybe use the text: 'TODO: Unpick'

src/components/Comment/Comment.stories.tsx
```diff
@@ -36,11 +36,36 @@ const commentData: CommentType = {
...
+export const Default = () => (
+  <Comment
+    comment={commentData}
+    pillar={"sport"}
+    setCommentBeingRepliedTo={() => {}}
+  />
+);
+Default.story = { name: "default" };
+
+export const StaffUser = () => (
```
oliverlloyd: Thanks for the story!

src/components/Comment/Comment.tsx
```diff
@@ -116,15 +118,41 @@ const Row = ({ children }: { children: JSX.Element | JSX.Element[] }) => (
 export const Comment = ({
   comment,
   pillar,
-  setCommentBeingRepliedTo
+  setCommentBeingRepliedTo,
+  user
 }: Props) => {
   const commentControlsButtonStyles = commentControlsButton(pillar);
+  const [scopedComment, setScopedComment] = useState(comment);
```
oliverlloyd: I don't think we should be loading the entire comment into state. This adds a lot of complexity to this component for a very limited benefit. It gives internal staff users immediate feedback on their actions but makes the codebase more complicated to reason about and increases the risk of bugs. Frontend doesn't have this feature - you only see updates after you refresh - so we have no requirement to actually add it

src/lib/api.tsx
```diff
@@ -152,12 +155,9 @@ export const getCommentCount = (
...
 };
 
-export {
-  getDiscussion,
-  getProfile,
-  preview,
-  comment,
-  reply,
-  recommend,
-  getPicks
```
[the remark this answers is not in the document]
liywjl: Good point, I will make it in a seperate PR. The issue I had we we were mixing `export const xxx` and 
```
export {
   xxx,
}
```

but granted, this should be outside the scope of this PR

src/components/Comment/Comment.tsx
liywjl: great catch, super easy change to implement both

src/components/Comment/Comment.tsx
```diff
@@ -119,11 +121,46 @@ export const Comment = ({
   comment,
   pillar,
   setCommentBeingRepliedTo,
+  user,
   isReply
 }: Props) => {
   const commentControlsButtonStyles = commentControlsButton(pillar);
+  const [isHighlighted, setIsHighlighted] = useState<boolean>(
+    comment.isHighlighted
+  );
```
oliverlloyd: Nice middle ground 👍

src/components/Comment/Comment.stories.tsx
```diff
@@ -56,16 +56,39 @@ export const Default = () => (
...
+export const StaffUserUnpicked = () => (
+  <Comment
+    comment={commentData}
+    pillar={"sport"}
+    setCommentBeingRepliedTo={() => {}}
+    user={staffUser}
+    isReply={false}
+  />
+);
+StaffUserUnpicked.story = { name: "Staff user unpicked" };
```
oliverlloyd: 💯

src/components/Comment/Comment.tsx
```diff
@@ -151,46 +162,53 @@ export const Comment = ({
...
+                  >
+                    {comment.userProfile.displayName}
+                  </a>
                 </div>
                 <div className={timestampWrapperStyles}>
                   <Timestamp
-                    isoDateTime={scopedComment.isoDateTime}
-                    linkTo={`https://discussion.code.dev-theguardian.com/comment-permalink/${scopedComment.id}`}
+                    isoDateTime={comment.isoDateTime}
+                    linkTo={`https://discussion.code.dev-theguardian.com/comment-permalink/${comment.id}`}
```
oliverlloyd: Not an issue for this PR but we'll need to be careful where we've got hardcoded refs to both production and code existing at the same time. We'll need to think about either sticking with prod or abstracting the baseUrl

PR #75, opened 2020-03-02
Ask username for first comment

liywjl:
## What does this change?

When the user posts their first comment, they should be asked to create a user name with a welcome text. They should be prevented for using the posting a comment until a user name is selected.

We use the users actual first and last name, in the welcome message, and we can extract this information from cookies, though we will likely wish to extract it from the root app.

## Why?

Parity with dotcom

https://trello.com/c/09kdF5rZ/1179-create-username-form-first-comment

### Current UI
Screenshot 2020-03-03 at 14 17 49

### Note
It is likely design will want some changes, so this may need to be revisited.


src/components/CommentForm/CommentForm.tsx
```diff
@@ -233,8 +233,32 @@ export const CommentForm = ({
...
+    if (response.status === "ok") {
+      // If we are able to submit userName we should continue with submitting comment
+      submitForm();
+      setFirstPost(false);
+    } else {
+      response.errors && setError(response.errors[0].message);
+    }
+  };
+
+  if (firstPost && body) {
```
oliverlloyd: Can we rename `firstPost` to `usernameMissing`? I think that will read a lot better

src/components/FirstCommentWelcome/FirstCommentWelcome.tsx
```diff
@@ -1,42 +1,166 @@
...
+            Username:{" "}
+          </span>
+          Must be 6-20 characters, letters and/or numbers only, no spaces.
+        </label>
+        <input
+          className={inputStyles}
+          type="text"
+          value={userName}
+          onChange={e => setUserName(e.target.value)}
+        />
```
oliverlloyd: We should see about using the design system for this input

liywjl: Currently under discussion with design, will come back to this once we have design's input

oliverlloyd:
Based on discussion with Akemi and the team, shall we use source foundation for the input (because it's better than custom elements and because it doesn't have rounded corners and then look to design to rework the form as required based on the result?

PR #78, opened 2020-03-03
Blocked comment

liywjl:
## What does this change?
When a comment is blocked, we need to change its styles and hide action buttons

## Why?
Parity with frontent

https://trello.com/c/HqP6FAun/1184-comment-was-removed-text-styles

### UI
Screenshot 2020-03-03 at 16 47 11

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { css } from "emotion";
+import { css, cx } from "emotion";
 
 import { neutral, space, palette } from "@guardian/src-foundations";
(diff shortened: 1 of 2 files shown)


src/components/Comment/Comment.tsx
```diff
@@ -130,6 +146,8 @@ export const Comment = ({
   );
   const [error, setError] = useState<string>();
 
+  const isCommentBlocked = comment.status === "blocked";
+
```
oliverlloyd: I reckon we could probably inline `comment.status === "blocked" ? ()` - fewer abstractions is always nice and it solves the !isTrue confusion

PR #88, opened 2020-03-06
Fix for firefox background color default

liywjl:
## What does this change?
Set the background of the comment action buttons 

## Why?
In firefox the background colour defaults to grey

## Before
Screenshot 2020-03-06 at 13 32 53

## After
Screenshot 2020-03-06 at 13 32 41

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -2,7 +2,7 @@ import React, { useState } from "react";
 import { css, cx } from "emotion";
 
 import { space, palette } from "@guardian/src-foundations";
-import { neutral } from "@guardian/src-foundations/";
+import { neutral } from "@guardian/src-foundations/palette";
 import { textSans } from "@guardian/src-foundations/typography";
 
 import { Pillar, CommentType, UserProfile } from "../../types";
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -1,7 +1,8 @@
 import React, { useState } from "react";
 import { css, cx } from "emotion";
 
-import { neutral, space, palette } from "@guardian/src-foundations";
+import { space, palette } from "@guardian/src-foundations";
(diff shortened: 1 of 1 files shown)


src/components/Comment/Comment.tsx
```diff
@@ -31,6 +32,7 @@ const commentControlsButton = (pillar: Pillar) => css`
   ${textSans.xsmall({ fontWeight: "bold" })}
   margin-right: ${space[2]}px;
   color: ${palette[pillar][400]};
+  background-color: ${neutral[100]};
```
oliverlloyd: The correct background colour is `background.primary` - we don't need to import `neutral` here

src/components/Comment/Comment.tsx
```diff
@@ -1,7 +1,8 @@
 import React, { useState } from "react";
 import { css, cx } from "emotion";
 
-import { neutral, space, palette } from "@guardian/src-foundations";
+import { space, palette } from "@guardian/src-foundations";
+import { neutral } from "@guardian/src-foundations/";
```
oliverlloyd: You should import from `@guardian/src-foundations/palette` - the fact you can import from the root is a bug

src/components/Comment/Comment.tsx
```diff
@@ -31,6 +32,7 @@ const commentControlsButton = (pillar: Pillar) => css`
   ${textSans.xsmall({ fontWeight: "bold" })}
   margin-right: ${space[2]}px;
   color: ${palette[pillar][400]};
```
oliverlloyd: Is it worth fixes these as we go along?

liywjl: prob best to have its own PR for most of these issues

src/components/Comment/Comment.tsx
liywjl: good catch

PR #93, opened 2020-03-14
Add list item to increase accessibilty

liywjl:
## What does this change?
Uses ul and li html tags for comments

## Why?
Increase accessibility for screen readers

https://trello.com/c/zmOLaFp7/1270-accessibility-use-ul-li-for-comment-list-and-top-picks-list

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -214,17 +214,19 @@ export const App = ({ shortUrl, user }: Props) => {
         ) : (
           <ul className={commentContainerStyles}>
             {comments.slice(0, 2).map(comment => (
-              <CommentContainer
-                key={comment.id}
-                comment={comment}
-                pillar="news"
-                shortUrl={shortUrl}
-                onAddComment={onAddComment}
-                user={user}
-                threads={filters.threads}
-                commentBeingRepliedTo={commentBeingRepliedTo}
-                setCommentBeingRepliedTo={setCommentBeingRepliedTo}
-              />
+              <li>
+                <CommentContainer
+                  key={comment.id}
+                  comment={comment}
+                  pillar="news"
+                  shortUrl={shortUrl}
+                  onAddComment={onAddComment}
+                  user={user}
+                  threads={filters.threads}
+                  commentBeingRepliedTo={commentBeingRepliedTo}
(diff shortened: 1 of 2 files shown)


src/components/CommentContainer/CommentContainer.tsx
```diff
@@ -129,15 +134,19 @@ export const CommentContainer = ({
...
-                pillar={pillar}
-                setCommentBeingRepliedTo={setCommentBeingRepliedTo}
-                user={user}
-                isReply={true}
-              />
-            ))}
+            <ul className={commentContainerStyles}>
+              {responses.map(responseComment => (
+                <li>
+                  <Comment
```
oliverlloyd: As above, how do you feel about making `Comment` an `li`?

src/App.tsx
```diff
@@ -286,17 +288,19 @@ export const App = ({ shortUrl, user }: Props) => {
...
-              shortUrl={shortUrl}
-              onAddComment={onAddComment}
-              user={user}
-              threads={filters.threads}
-              commentBeingRepliedTo={commentBeingRepliedTo}
-              setCommentBeingRepliedTo={setCommentBeingRepliedTo}
-            />
+            <li>
+              <CommentContainer
+                key={comment.id}
```
oliverlloyd: Another reason to not have the wrapper is this `key` prop won't need to be lifted

src/components/CommentContainer/CommentContainer.tsx
```diff
@@ -117,7 +122,7 @@ export const CommentContainer = ({
   };
 
   return (
-    <div>
+    <>
```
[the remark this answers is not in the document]
liywjl: I think it helps to reason with when the `ul` and `li` are in the same file, sure we can remove a fragment, but I feel that a fragment here isnt much of an issue in accessibility/performance/readability. To the contrary I think it improves readability

src/App.tsx
liywjl: good catch :)

src/components/CommentContainer/CommentContainer.tsx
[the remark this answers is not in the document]
oliverlloyd: Fair enough. Happy to merge 👍

PR #94, opened 2020-03-14
Dropdown

oliverlloyd:
## What does this change?
This adds a custom Dropdown component so we don't need to depend on browser based styling for select fields

```typescript
type DropdownLinkType = {
  value: string;
  title: string;
  isActive?: boolean;
};

type Props = {
  id: string;
  label: string;
  links: DropdownLinkType[];
  pillar: Pillar;
  onSelect: (value: string) => void;
};
```

### With first item selected
Screenshot 2020-03-14 at 15 00 04

### With nothing selected
Screenshot 2020-03-14 at 14 31 25

### With narrower content
Screenshot 2020-03-14 at 14 59 54

## Why?
There's an argument for keeping classic, simple select styles based on the OS/Browser but the [designs for the filters](https://www.figma.com/file/55MFs13FCdCIfkhcgId4bh/Comment-filter-pagination?node-id=1%3A7) need a more compact look

https://trello.com/c/7V2UJwGq/1231-style-filters

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,6 @@
     "@guardian/src-button": "^0.8.0",
     "@guardian/src-foundations": "^0.12.3",
     "@guardian/src-svgs": "^0.1.0",
-    "@testing-library/jest-dom": "^4.2.4",
-    "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
     "@types/jest": "^25.1.0",
     "@types/node": "^13.5.0",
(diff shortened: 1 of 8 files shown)


SiAdcock:
Oooh, tasty reusable components! I feel like a kid in a candy shop 🍬

oliverlloyd:
> Oooh, tasty reusable components! I feel like a kid in a candy shop 🍬

There are actually 2 flavours of dropdown components. Those with links (think edition) and those with callbacks (as used here). The link version could be built sans javascript, something I'm interested in but this callback version not so much.

The 2 versions seemed different enough for me to branch so I did, meaning the discussion and DCR `Dropdowns` are not the same but you could merge the 2 use cases into one shared component if you don't object to complex, mixed flavours

SiAdcock:
Interesting point on the flavours, thanks for sharing. We have a similar distinction with buttons (buttons for non-navigational actions and buttons that act as links). To me, callback dropdown (in this case filters) and navigational (link) dropdowns sound like different components, at least from the code’s perspective

PR #102, opened 2020-03-19
Comment Loading

oliverlloyd:
## What does this change?
Adds a new component `LoadingComments` for showing a loading skeleton for a comment

## Also included: Grey
A helper component for building loading skeletons - this will likely get broken out into it's own component soon

```typescript
type Props = {
  height: string;
  width?: string;
  spaceBelow?: 1 | 2 | 3 | 4 | 5 | 6 | 9;
  spaceLeft?: 1 | 2 | 3 | 4 | 5 | 6 | 9;
}
```
## Why?
This is in preparation for showing loading state

https://trello.com/c/PbTX8ybF/1213-loading-state-in-comments

### DIFF
--- a/src/components/LoadingComments/LoadingComments.stories.tsx
+++ b/src/components/LoadingComments/LoadingComments.stories.tsx
@@ -3,11 +3,5 @@ import { LoadingComments } from "./LoadingComments";
 
 export default { component: LoadingComments, title: "LoadingComments" };
 
-export const Default = () => {
-  return (
-export const Default = () => <LoadingComments />
-      <LoadingComments />
-    </>
-  );
-};
+export const Default = () => <LoadingComments />;
 Default.story = { name: "default" };
--- a/src/components/LoadingComments/LoadingComments.stories.tsx
+++ b/src/components/LoadingComments/LoadingComments.stories.tsx
@@ -1,0 +1,13 @@
+import React from "react";
+import { LoadingComments } from "./LoadingComments";
+
+export default { component: LoadingComments, title: "LoadingComments" };
+
+export const Default = () => {
+  return (
+    <>
+      <LoadingComments />
+    </>
+  );
+};
+Default.story = { name: "default" };
--- a/src/components/LoadingComments/LoadingComments.tsx
+++ b/src/components/LoadingComments/LoadingComments.tsx
@@ -8,10 +8,10 @@ const BACKGROUND_COLOUR = neutral[93];
 
 const shimmer = keyframes`
   0% {
-    background-position: -1000px 0;
+    background-position: -1500px 0;
   }
   100% {
-    background-position: 1000px 0;
+    background-position: 1500px 0;
   }
 `;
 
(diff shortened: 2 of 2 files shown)


src/components/LoadingComments/LoadingComments.stories.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { LoadingComments } from "./LoadingComments";
+
+export default { component: LoadingComments, title: "LoadingComments" };
+
+export const Default = () => {
+  return (
+    <>
```
liywjl: no need for fragments
```suggestion
export const Default = () => <LoadingComments />
```

src/components/LoadingComments/LoadingComments.stories.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { LoadingComments } from "./LoadingComments";
+
+export default { component: LoadingComments, title: "LoadingComments" };
+
+export const Default = () => {
+  return (
+export const Default = () => <LoadingComments />
+      <LoadingComments />
```
liywjl: hehe you might need to remove this and the closing fragments as well :P  sorry

src/components/LoadingComments/LoadingComments.tsx
```diff
@@ -0,0 +1,113 @@
...
+  margin-right: ${space[2]}px;
+
+  ${shimmerStyles}
+`;
+
+const Column = ({ children }: { children: JSX.Element | JSX.Element[] }) => (
+  <div
+    className={css`
+      display: flex;
+      flex-direction: column;
```
liywjl: the common use of:
```
display: flex;
flex-direction: column;
```
 and 
```
display: flex;
flex-direction: row;
```
makes me think we should starting thing of some sort of abstraction to avoid duplicate CSS

src/components/LoadingComments/LoadingComments.stories.tsx
oliverlloyd: No worries, done

src/components/LoadingComments/LoadingComments.tsx
```diff
@@ -0,0 +1,113 @@
...
+  />
+);
+
+export const LoadingComments = () => (
+  <>
+    <div className={containerStyles}>
+      <div className={avatarStyles(48)} />
+      <Column>
+        <Row>
+          <Grey height="20px" width="140px" spaceBelow={9} />
```
liywjl: I have mixed feeling about mixing numbers with strings ending with `px`, it should either be one or the other

src/components/LoadingComments/LoadingComments.tsx
liywjl: worth making bringing this to src-foundations? as a grid component?

oliverlloyd: Yeah, tempting. The problem is we're not using `styled` so it'a harder to compose components. It'd be great to do: `const PinkRow = styled(Row)'color: pink';` but if we can't then you end up adding lots of divs or creating the propocalypse by adding new props for every new use case 😱

PR #105, opened 2020-03-20
baseUrl

oliverlloyd:
## What does this change?
Adds support for passing in a `baseUrl` prop to `App` so that we accept the domain to make requests against dynamically

## Why?
So that we can easily test over CODE and production

https://trello.com/c/fNn0CFdy/1310-pass-in-a-baseurl-to-comments

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -120,7 +120,7 @@ const readFiltersFromLocalStorage = (): FilterOptions => {
   };
 };
 
-export const App = ({ shortUrl, user, additionalHeaders }: Props) => {
+export const App = ({ baseUrl, shortUrl, user, additionalHeaders }: Props) => {
   const [filters, setFilters] = useState<FilterOptions>(
     readFiltersFromLocalStorage()
   );
@@ -208,7 +208,7 @@ export const App = ({ shortUrl, user, additionalHeaders }: Props) => {
     setComments([simulateNewComment(commentId, body, user), ...comments]);
   };
 
-  initialiseApi({ additionalHeaders });
+  initialiseApi({ additionalHeaders, baseUrl });
 
   const showPagination = totalPages > 1;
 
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -17,7 +17,7 @@ import {
   getDiscussion,
   getCommentCount,
   getPicks,
-  setAdditionalHeaders
+  initialiseApi
```
oliverlloyd: I change this function to be more generic so it can also set the baseUrl

src/components/Comment/Comment.tsx
```diff
@@ -227,7 +234,15 @@ export const Comment = ({
                 <div className={timestampWrapperStyles}>
                   <Timestamp
                     isoDateTime={comment.isoDateTime}
-                    linkTo={`https://discussion.theguardian.com/comment-permalink/${comment.id}`}
+                    linkTo={joinUrl([
+                      // Remove the discussion-api path from the baseUrl
+                      baseUrl
+                        .split("/")
+                        .filter(path => path !== "discussion-api")
+                        .join("/"),
```
oliverlloyd: Annoyingly, the base url required for building permalink urls does not include the `discussion-api` in the path, even though this is part of the 'base' url. So here I'm removing that part of the path

src/lib/api.tsx
```diff
@@ -11,13 +11,22 @@ import {
   AdditionalHeadersType
 } from "../types";
 
-const baseURL = "https://discussion.theguardian.com/discussion-api";
-
-let additionalHeaders = {};
+let options = {
+  // Defaults
+  baseUrl: "https://discussion.theguardian.com/discussion-api",
```
oliverlloyd: I've set this to default to production but in reality the prop is required so this should never be needed

src/App.tsx
```diff
@@ -207,7 +208,7 @@ export const App = ({ shortUrl, user, additionalHeaders }: Props) => {
     setComments([simulateNewComment(commentId, body, user), ...comments]);
   };
 
-  setAdditionalHeaders(additionalHeaders);
+  initialiseApi({ additionalHeaders, baseUrl });
```
liywjl: nice :)

src/lib/api.tsx
```diff
@@ -42,19 +51,19 @@ export const getDiscussion = (
   };
   const params = objAsParams(apiOpts);
 
-  const url = joinUrl([baseURL, "discussion", shortUrl, params]);
+  const url = joinUrl([options.baseUrl, "discussion", shortUrl]) + params;
```
oliverlloyd: By passing `params` through `joinUrl` it was adding an extra '/' at the end of the path and the discussion api doesn't handle trailing slashes and returns a 404

PR #111, opened 2020-03-21
More consistent text sizes

oliverlloyd:
## What does this change?
Ensure we're only using 2 text sizes everywhere. `textSans.small` and `textSans`xsmall`

## Why?
To bring us in line with the design system and reduce font size disparity

## Before
Screenshot 2020-03-21 at 17 41 33

## After
Screenshot 2020-03-21 at 17 41 02

https://trello.com/c/kINjHtFy/1321-consistent-text-sizing

### DIFF
--- a/src/components/CommentForm/CommentForm.tsx
+++ b/src/components/CommentForm/CommentForm.tsx
@@ -79,6 +79,10 @@ const buttonContainerStyles = css`
   }
 `;
 
+const buttonTextStyles = css`
+  ${textSans.small({ fontWeight: "bold" })}
+`;
+
 const headerTextStyles = css`
   margin: 0;
   ${textSans.xsmall()};
@@ -387,7 +391,7 @@ export const CommentForm = ({
         <div className={bottomContainer}>
           <div className={buttonContainerStyles}>
             <Button type="submit" size="small">
-              Post your comment
+              <div className={buttonTextStyles}>Post your comment</div>
             </Button>
             {(isActive || body) && (
               <>
@@ -396,10 +400,10 @@ export const CommentForm = ({
                   onClick={fetchShowPreview}
                   priority="secondary"
                 >
-                  Preview
+                  <div className={buttonTextStyles}>Preview</div>
                 </Button>
                 <Button size="small" onClick={resetForm} priority="tertiary">
-                  Cancel
+                  <div className={buttonTextStyles}>Cancel</div>
                 </Button>
               </>
             )}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,7 +4,7 @@ import { css } from "emotion";
 import "regenerator-runtime/runtime";
 
 import { Button } from "@guardian/src-button";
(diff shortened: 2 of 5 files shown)


src/components/CommentForm/CommentForm.tsx
```diff
@@ -387,7 +391,7 @@ export const CommentForm = ({
         <div className={bottomContainer}>
           <div className={buttonContainerStyles}>
             <Button type="submit" size="small">
-              Post your comment
+              <div className={buttonTextStyles}>Post your comment</div>
```
liywjl: do we need to use a div? could the styles not be applied to the button?

src/components/CommentForm/CommentForm.tsx
```diff
@@ -396,10 +400,10 @@ export const CommentForm = ({
                   onClick={fetchShowPreview}
                   priority="secondary"
                 >
-                  Preview
+                  <div className={buttonTextStyles}>Preview</div>
```
liywjl: same here

src/components/CommentForm/CommentForm.tsx
```diff
@@ -396,10 +400,10 @@ export const CommentForm = ({
...
                 </Button>
                 <Button size="small" onClick={resetForm} priority="tertiary">
-                  Cancel
+                  <div className={buttonTextStyles}>Cancel</div>
```
liywjl: same here

src/components/CommentForm/CommentForm.tsx
SiAdcock: You can [override styles](https://www.theguardian.design/2a1e5182b/p/2619e3-overriding-styles) of Source components. It would be useful to understand why we need to in this case... is it just for visual parity, or is there a presentational reason?

oliverlloyd: It's for presentational reasons. The button text is larger than the standard text size used throughput discussion (textSans.small) and seemed a bit, well, big. But I'm happy to take this out and stick with src. I don't have strong opinions here.

## Before
Screenshot 2020-03-21 at 22 47 15

## After
Screenshot 2020-03-21 at 22 47 02

oliverlloyd: Re: passing classNames. It's not my favourite thing to do because it abstracts the styling in a way that doesn't scale and creates this strange coupling where I lose control over what styles might win in a war. If I wrap it in a div then I control things. But it's an extra div 🤷‍♂

SiAdcock: Right on. I'll discuss the text size thing with Akemi. It's certainly something we should keep an eye on.

Regarding the extra div, I don't mind as long as it still gets read out by screenreaders 🙂

oliverlloyd: > as long as it still gets read out by screenreaders

Good call. I tested using a screenreader and it works fine 👍

PR #112, opened 2020-03-21
Limit pick text

oliverlloyd:
## What does this change?
If the comment text of a pick is especially long we now only show the first 450 chars

## Before
Screenshot 2020-03-21 at 20 10 07

## After
Screenshot 2020-03-21 at 20 08 42

## Why?
Because sometimes they really are quite long and we don't want to push all the content down too far because of that

https://trello.com/c/49rX9wbP/1323-limit-pick-text

### DIFF
--- a/src/components/TopPick/TopPick.tsx
+++ b/src/components/TopPick/TopPick.tsx
@@ -80,6 +80,12 @@ const linkStyles = css`
   }
 `;
 
+const truncateText = (input: string, limit: number) => {
+  // If input greater than limit trucate by limit and append an ellipsis
+  if (input.length > limit) return input.substr(0, limit) + "&#8230;";
+  return input;
+};
+
 export const TopPick = ({
   baseUrl,
   comment
@@ -98,7 +104,9 @@ export const TopPick = ({
       >
         Guardian Pick
       </h3>
-      <p dangerouslySetInnerHTML={{ __html: comment.body }}></p>
+      <p
+        dangerouslySetInnerHTML={{ __html: truncateText(comment.body, 450) }}
+      ></p>
     </div>
     <div className={pickMetaWrapper}>
       <div className={userDetails}>
--- a/src/components/TopPick/TopPick.tsx
+++ b/src/components/TopPick/TopPick.tsx
@@ -38,7 +38,7 @@ const pickComment = css`
 
   :before {
     content: "";
-    margin-left: 24px;
+    margin-left: ${space[6]}px;
     position: absolute;
     border-right: ${arrowSize}px solid transparent;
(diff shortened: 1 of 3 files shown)


SiAdcock:
Looks great!

Small UX thing: Do you think it would be better to break at the end of a word, rather than at exactly 450 characters, which might be in the middle of a word?

src/components/TopPick/TopPick.tsx
```diff
@@ -0,0 +1,158 @@
...
+`;
+
+const userMetaStyles = css`
+  display: flex;
+  flex-direction: column;
+`;
+
+const userName = css`
+  font-weight: bold;
+  color: ${palette.news.main}; /* TODO USE PILLAR */
```
SiAdcock: Could use news.main instead of palette

src/components/TopPick/TopPick.tsx
```diff
@@ -0,0 +1,158 @@
...
+  margin-bottom: ${arrowSize + 5}px;
+  position: relative;
+
+  ${from.tablet} {
+    min-height: 150px;
+  }
+
+  :before {
+    content: "";
+    margin-left: 24px;
```
SiAdcock: space[6]

oliverlloyd:
> Do you think it would be better to break at the end of a word, rather than at exactly 450 characters, which might be in the middle of a word?

I've not thought of doing that way before but it could make the break less sudden I guess. The advatage is the reader can complete reading the word but that won't help them in terms of knowing the end of the sentance so the benefit is small. There's an argument people are used to ellipsis breaks in these contexts so it shouldn't be too jarring and maybe breaking inside a word makes it clear that it is really a break, not the end of a sentance?

Another point is that frontend breaks inside the word so we will preserve parity by staying with 450 characters.

src/components/TopPick/TopPick.tsx
oliverlloyd: I've got another ticket to make this colouring dynamic based on `pillar` so we need the full palette

SiAdcock:
Yeah, all good points. I'll add a placeholder to discuss it with a UXer, I'm keen to enshrine this kind of thing in the design system so we're consistent everywhere. I'll link to this discussion as a useful starting point!

src/components/TopPick/TopPick.tsx
SiAdcock: You could create your own little pillar colour dictionary. I'm not sure how much of a saving you'll get over importing the entire palette though.

```js
const pillarColours = {
  news: news[400],
  sport: sport[400],
  // ...
}
```

PR #124, opened 2020-03-25
Comment mobile layout

liywjl:
## What does this change?
On mobile width now the image will appear as part of the comment header row rather than in its own column

## Why?
Parity with frontend

## Before
Screenshot 2020-03-25 at 14 36 25

Screenshot 2020-03-25 at 14 36 12

## After
Screenshot 2020-03-25 at 14 34 01

Screenshot 2020-03-25 at 14 34 23

https://trello.com/c/m6Td7ai3/1332-comment-mobile-layout


src/components/Comment/Comment.tsx
```diff
@@ -78,7 +71,7 @@ const commentWrapper = css`
   padding: ${space[2]}px 0;
 `;
 
-const avatarMargin = css`
+const marginRight = css`
```
oliverlloyd: 👍 this is a better name

src/components/Comment/Comment.tsx
```diff
@@ -88,28 +81,15 @@ const commentProfileName = (pillar: Pillar) => css`
   ${textSans.small({ fontWeight: "bold" })}
 `;
 
-const commentDetails = css`
+const flexGrow = css`
```
oliverlloyd: 👍

src/components/Comment/Comment.tsx
```diff
@@ -123,31 +103,31 @@ const flexRowStyles = css`
...
+const fullWidthStyles = css`
+  width: 100%;
+`
+
+const spaceBetweenStyles = css`
+  display: flex;
+  justify-content: space-between;
+`;
+
+const headerStyles = css`
```
oliverlloyd: Is this redundant if we already have `spaceBetweenStyles`?

src/components/Comment/Comment.tsx
```diff
@@ -207,116 +287,133 @@ export const Comment = ({
...
+                displayName={comment.userProfile.displayName}
+                size={isReply ? "small" : "medium"}
+              />
+            </div>
+            <div className={flexColStyles}>
+              <header className={cx(flexRowStyles, headerStyles, fullWidthStyles)}>
+                <div className={flexColStyles}>
+                  <div className={cx(flexRowStyles, alignItemsCenter)}>
+                    <div className={marginRight}>
+                      <ProfilName
```
oliverlloyd: I don't think there's any benefit in this abstraction. I'd be in favour of less DRY and more repetition to make the code more readable and maintainable.

src/components/Comment/Comment.tsx
```diff
@@ -207,116 +287,133 @@ export const Comment = ({
...
+                  {comment.status !== "blocked" && (
+                    <RecommendationCount
+                      commentId={comment.id}
+                      initialCount={comment.numRecommends}
+                      alreadyRecommended={false}
+                    />
+                  )}
+                </>
+              </header>
+              <CommentMessage
```
oliverlloyd: Maybe this new component could live in its own file?

src/components/Comment/Comment.tsx
```diff
@@ -207,116 +287,133 @@ export const Comment = ({
...
+                imageUrl={comment.userProfile.avatar}
+                displayName={comment.userProfile.displayName}
+                size={isReply ? "small" : "medium"}
+              />
+            </div>
+            <div className={flexColStyles}>
+              <header className={cx(flexRowStyles, headerStyles, fullWidthStyles)}>
+                <div className={flexColStyles}>
+                  <div className={cx(flexRowStyles, alignItemsCenter)}>
+                    <div className={marginRight}>
```
oliverlloyd: This is pretty hard to reason about. I much preferred how it was before with 

```
<Column>
  <Row>
    etc
```

If the goal is readability, I think the composition approach wins here

src/components/Comment/Comment.tsx
```diff
@@ -161,19 +141,56 @@ const ReplyArrow = () => (
...
+}: {
+  comment: CommentType,
+  pillar: Pillar,
+  setCommentBeingRepliedTo: (commentBeingRepliedTo?: CommentType) => void,
+  user?: UserProfile,
+  isHighlighted: boolean,
+  setIsHighlighted: (isHighlighted: boolean) => void,
+  setError: (errorMessage: string) => void
+}) => {
   const commentControlsButtonStyles = commentControlsButton(pillar);
```
oliverlloyd: I know this is outside this PR but we could maybe remove this abstraction?

oliverlloyd:
It'd also be great to see some stories for these layouts!

src/components/Comment/Comment.tsx
liywjl: I think `<Column>` and `<Row>` should be their own components and should be imported to avoid flexbox CSS, I am happy to do that here

src/components/Comment/Comment.tsx
liywjl: just missed this, thanks for pointing out

src/components/Comment/Comment.tsx
liywjl: good point

src/components/Comment/Comment.tsx
liywjl: good point, I hoped it would make it easier to read, but it doesnt add much

src/components/Comment/Comment.tsx
liywjl: good point

src/App.stories.tsx
```diff
@@ -3,24 +3,10 @@ import { App } from "./App";
 import { css } from "emotion";
 
 import { mockSignIn } from "./lib/mockSignIn";
+import { user as aUser } from "./fixtures/user";
```
oliverlloyd: As per my other comments, this abstraction might be useful in the future, possibly, but right now you could declare this mock locally to help keep the tests easy to reason through.

src/components/Column/Column.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { css, cx } from "emotion";
+
+const flexRowStyles = css`
+  display: flex;
+  flex-direction: column;
+`;
```
oliverlloyd: typo in var name, should be Column

src/components/Column/Column.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export const Column = ({
+  children,
+  className
+}: {
+  children: React.ReactNode;
+  className?: string;
```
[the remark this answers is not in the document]
liywjl: it may be a common pattern in libs, but I do not see it as a work around. I feel that if it does exactly what it says then it is an ok abstraction.

My vision for `Column` and `Row` is to mimic `div` HTML elements, for the moment the only prop we would need is classNames, but we could also support useRef's and other props in the future.

I also do not fully understand what you are referring to as an alternative solution, I am guessing you mean we should just have styles applied in a child wrapper? ex:
Currently:
```
<Row className={someNewClass}>
    ...
</Row>
```

Your suggestion:
```
<Row>
   <div className={someNewClass}>
    ...
    </div>
</Row>
```

Feel free to correct me if i misunderstand.

I would be interested in getting @gtrufitt take on this approach

src/components/Row/Row.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { css, cx } from "emotion";
+
+const flexRowStyles = css`
+  display: flex;
+  flex-direction: row;
+`;
+
+export const Row = ({
+  children,
+  className = ""
```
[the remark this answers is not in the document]
liywjl: we should keep the conversation in this thread https://github.com/guardian/discussion-rendering/pull/124#discussion_r398957194

src/components/CommentMessage/CommentMessage.stories.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+
+import { CommentMessage } from "./CommentMessage";
+import { comment } from "../../fixtures/comment";
+import { staffUser } from "../../fixtures/user";
```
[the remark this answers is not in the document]
liywjl: the abstraction originated because I needed a user, and saw one defined in `App.tsx`. 
I noticed that there are variations between staff and non staff users, so I think this abstraction is merited now.

src/components/Column/Column.tsx
liywjl: good catch! thx

src/components/Column/Column.tsx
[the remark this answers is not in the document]
oliverlloyd: So the issue with passing 'className' as a prop is it essentially lets you do anything. It's basically the `any` of props and as such should be used as a last resort when you have no other choice.

But we have other choices, we control the code. The problem here is you want to make `Row` work for multiple use cases with different styles.
So we could:

1. Have style specific props, like `isFullWidth` or `backgroundColor` etc.
2. Have different components, like `FullWidthRow` or `BlueRow`
3. Use `styled` like `const BlueRow = styled(Row)`background: blue`
4. Keep `Row` basic and have a second div with the context specific styles applied to it
5. Not abstract at all, just define the components needed locally

My preference would be to have a basic `Row` component if there are several places in the code that need it and there's a use case for it. But where the requirements are to have different styles applied then we should manage them them locally, either by extra divs or by just adding the css needed as normal, not using Row because Row isn't what we want in that situation.

We've hit this situation multiple times in the code here and on DCR, where a seemingly generic component turned out to have subtle use cases needing extra styles and in all those situations we've not resorted to `className` (except for Adslot) and I'm super reticent to start adding this pattern now.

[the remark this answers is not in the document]
oliverlloyd: To be clear, my dislike of the `className` pattern is that it does not scale. The whole point of react and typescript is to keep code maintainable as it scales but when you let people start passing classNames around you lose control and readability and your code rapidly turns into spaghetti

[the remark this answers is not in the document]
gtrufitt: Hmm, I guess I'm slightly split in this context - having `className` on layout elements seems like a reasonable thing and I've certainly wished for it before rather than wrapping elements.

However, on balance, I agree with Oliver purely for the reason that it keeps the layout components clear in their responsibility, and if we think that they need extra functionality passing it as props is a nicer UX as auto-completable. 

It also keeps styling that we might otherwise be tempted to pass through, creating a mis-direction, close to the component that's using it.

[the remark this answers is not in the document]
gtrufitt: Ok yeah, Oliver replied as I was typing, he's a faster typer.

src/components/CommentMessage/CommentMessage.stories.tsx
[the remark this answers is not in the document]
oliverlloyd: The problem though is it's only used in one place. So all you've achieved is to take some code from that file and put it in a different one. Now when someone comes to this file to learn about or fix a problem with this test they will need to context switch between two different files, which has a cognitive cost.

We should be slow to do things that increase this cost and only do them when there's benefit that makes the move worth it. That benefit could be if the file is getting excessively large requiring lots of scrolling or if the same code is shared in many places and, crucially, that you are confident that changes to that shared code should always be applied in all the places it is used.

src/components/Column/Column.tsx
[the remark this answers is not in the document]
oliverlloyd: To give some context for my passion on this point. I worked on a project that used this pattern and I grew to Utterly Hate It. I'd be constantly tracing `className` props though the code to try and see where styles were coming from and repeatedly ended up lost and frustrated. My stress levels rose significantly which may or may not have shortened my lifespan. All because of `className` props.

I spent a large part of my time refactoring it away which was slow and painful because you had to relearn all the use cases and break the abstractions.

I _get_ that it's tempting. It's tempting because it's easy, as soon as you add it all your problems go away and life is great. But it's a pact with the devil, you are deferring the pain for later, probably to someone else, maybe me, and I'm don't want to go through that ever again.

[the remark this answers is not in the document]
liywjl: @oliverlloyd I am coming around to the idea of keeping both components minimalist.

I have another suggestion., maybe we could make the `Row` and `Column` just a flex-box abstraction? By adding common flex-box options the we use `justify-content: space-between;` and `width: 100%`

Or would this go counter to your point?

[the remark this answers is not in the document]
oliverlloyd: My *final* point: Why is this okay for styles but not for other types of prop? if you think about it, it's the same as just having `randomProps`  If you're going to say `className` you might as well call it `someRandomStylesWhahayLetsGo` and then you can call the other props `randomPropsYeah` and just spread that untyped prop love everywhere.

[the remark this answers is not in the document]
oliverlloyd: > By adding common flex-box options the we use `justify-content: space-between;` and `width: 100%`

Yep, sure. This makes sense and is readable. I have zero issues with this. 👍 

There is that classic risk of reimplementing the flex api - most people have been there - but I'm pretty sure if you open most mature react codebases you will find similar components and there's nothing wrong with them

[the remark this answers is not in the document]
liywjl: This has been a great discussion, thanks for sharing @oliverlloyd. 

I would love to have a similar abstraction in dotcom-rendering.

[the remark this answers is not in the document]
gtrufitt: Thanks for this, really interesting. I guess DCR has https://github.com/guardian/dotcom-rendering/blob/master/src/web/components/Flex.tsx

PR #126, opened 2020-03-26
Show replies for permalinks

oliverlloyd:
## What does this change?
This ensures that reply threads are expanded when a permalink is being used

## Why?
Because in "collapsed" mode some replies are hidden and can't be scrolled

## Depends on
https://github.com/guardian/dotcom-rendering/pull/1298

https://trello.com/c/oD8wVrK2/1345-permalink-to-hidden-replies

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,7 +13,8 @@ import {
   UserProfile,
   AdditionalHeadersType,
   PageSizeType,
-  OrderByType
+  OrderByType,
+  ThreadsType
 } from "./types";
 import {
   getDiscussion,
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -102,16 +103,25 @@ const rememberFilters = (filtersToRemember: FilterOptions) => {
...
+}) => {
   const initialisedFilters = initFiltersFromLocalStorage();
   return {
     ...initialisedFilters,
     // Override if prop given
     pageSize: pageSizeOverride || initialisedFilters.pageSize,
-    orderBy: orderByOverride || initialisedFilters.orderBy
+    orderBy: orderByOverride || initialisedFilters.orderBy,
+    threads:
+      initialisedFilters.threads === "collapsed" && permalinkBeingUsed
```
liywjl: wouldn't this  need to be wrapped like : `(initialisedFilters.threads === "collapsed" && permalinkBeingUsed)` to avoid false being set?

oliverlloyd: I think because it's using the '?' ternary it will alway return one of the 2 options and `initialisedFilters.threads` can't be undefined so we should be good.

liywjl: my understanding is if: `initialisedFilters.threads === "collapsed" ` is false, then  `&& permalinkBeingUsed` is never called, therefore setting `threads` to false as it never reaches the `?`. but wrapping it with `( ... )` forces the program to evaluate it as 1 condition

oliverlloyd: Ternaries always return one or other of their outcomes so if threads is false it will jump to the negative outcome

Screenshot 2020-03-26 at 12 20 58

PR #129, opened 2020-03-26
Optimistic replies

liywjl:
## What does this change?
Add optimistic reply rendering

## Why?
We have this logic already for comments, for consistency this logic should apply to replies

https://trello.com/c/pBVxrFs9/1340-optimistic-replies

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -232,7 +232,12 @@ export const App = ({
     setPage(page);
   };
 
-  const onAddComment = (commentId: number, body: string, user: UserProfile, commentBeingRepliedTo?: CommentType) => {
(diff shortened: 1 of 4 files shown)


liywjl:
> I was struggling to reason through this code for a while when I finally realised the reason it's so complex is we both missed the fact that `CommentContainer` holds the `responses` in state! You don't need to use onAddComment or pass around indexes to edit the responses for the comment, you can simply add a simulated reply to the responses array inside `CommentContainer` 🎉

You are right! that would greatly simplify this PR. Thank you for pointing this out! I will close this PR and reopen on with this new simplified approch.

PR #130, opened 2020-03-26
Reply links to login page when not logged in

oliverlloyd:
## What does this change?
This make the reply button a link to the login page for readers that are not logged in

## Why?
Because they can't reply and need to login

https://trello.com/c/vhCuq6wM/1351-hide-reply-action-when-not-logged-in


oliverlloyd:
> On frontend we still display the reply button but just redirect the user to login

Ah, I never noticed that - good spot!

src/components/Comment/Comment.tsx
```diff
@@ -119,8 +119,9 @@ const timestampWrapperStyles = css`
   justify-content: center;
 `;
 
-const linkStyles = css`
-  color: inherit;
+const linkStyles = (pillar: Pillar) => css`
+${textSans.xsmall({ fontWeight: "bold" })}
```
liywjl: strange formattting

oliverlloyd: Excellent spot! There was a missing semi-colon. Fixed now

PR #134, opened 2020-03-28
Use pillar for colours

oliverlloyd:
## What does this change?
Adds a `pillar` prop to App and passes it down so it can be used to colour elements

### E.g. Sport
Screenshot 2020-03-27 at 22 45 23
### E.g. Opinion
Screenshot 2020-03-27 at 22 45 42

## Why?
So that discussion follows the same colour palette as used elsewhere

https://trello.com/c/uSEmtQWg/1318-use-pillar-for-colours


src/components/TopPick/TopPick.tsx
```diff
@@ -67,9 +68,9 @@ const userMetaStyles = css`
   flex-direction: column;
 `;
 
-const userName = css`
+const userName = (pillar: Pillar) => css`
   font-weight: bold;
-  color: ${palette.news.main}; /* TODO USE PILLAR */
+  color: ${palette[pillar].main}; /* TODO USE PILLAR */
```
liywjl: you can remove the TODO here

PR #138, opened 2020-03-29
Prevent Page Zero

oliverlloyd:
## What does this change?
This corrects the start index in pagination being too low

## Before
0 to 25, 25 - 50, 50 - 75

## After
1 to 25, 26 - 50, 51 - 75

## Why?
Because we want these numbers to read naturally

https://trello.com/c/BitwNOL5/1365-page-zero


src/components/Pagination/Pagination.tsx
```diff
@@ -217,11 +217,10 @@ export const Pagination = ({
...
   // Pagination Text
-  const startIndex = filters.pageSize * (currentPage - 1);
-  const endIndex =
-    filters.pageSize * currentPage < commentCount
-      ? filters.pageSize * currentPage
-      : commentCount;
+  let { pageSize } = filters;
+  let endIndex = pageSize * currentPage;
+  let startIndex = endIndex - (pageSize - 1);
+  if (endIndex > commentCount) endIndex = commentCount;
```
gtrufitt: I really don't know if this is helpful or not, but the example helps me reason with the code a little quicker. 

```suggestion
  let { pageSize } = filters; // e.g: pageSize of 25
  let endIndex = pageSize * currentPage; // e.g: currentPage is page 2, endIndex is 25 * 2 = 50
  let startIndex = endIndex - (pageSize - 1); // e.g: 50 - (25 - 1) startIndex is 26. Example page 2 has comments 26 - 50 on it.
  if (endIndex > commentCount) endIndex = commentCount; // If there are less total comments than allowed on the page, endIndex is total comment count
```

PR #139, opened 2020-03-29
Refactor pagination

oliverlloyd:
## What does this change?
This refactors pagination to fix a couple of bugs

1. If the reader changes page size up this can leave them on an invalid page, causing api errors
2. The total page size count was incorrect for `expanded` and `collapsed` views. Before we were using a separate call to get the count but that was returning the wrong value in many cases. Instead we can use the `topLevelCommentCount` property on the `getDiscussion` response. This isn't the real count of comments but it's the correct count of items in relation to how things are paginated, which is what we want. E.g. If you choose unthreaded, every comment counts as a pagination item and the commentCount goes up. If you choose threaded (expanded or collapsed) then only the top level comment is counted and the total count of items goes down.

https://trello.com/c/1yff38ws/1366-refactor-pagination


src/App.tsx
```diff
@@ -183,21 +183,12 @@ export const App = ({
       setLoading(false);
       if (json?.status !== "error") {
         setComments(json?.discussion?.comments);
+        setCommentCount(json.discussion.topLevelCommentCount);
```
gtrufitt: Above we use `?.` syntax, do we want to do the same?

src/App.tsx
```diff
@@ -237,6 +228,20 @@ export const App = ({
   }, [comments, commentToScrollTo]); // Add comments to deps so we rerun this effect when comments are loaded
 
   const onFilterChange = (newFilterObject: FilterOptions) => {
+    // If we're reducing the pageSize then we may need to override the page we're on to prevent making
+    // requests for pages that don't exist.
+    // E.g. If we used to have 102 comments and a pageSize of 25 then the current page could be 5 (showing 2
+    // comments). If we then change pageSize to be 50 then there is no longer a page 5 and trying to ask for it
+    // from the api would return an error so, in order to respect the readers desire to be on the last page, we
+    // need to work out the maximum page possible and use that instead.
+    let maxPagePossible = Math.floor(commentCount / newFilterObject.pageSize);
```
gtrufitt: Good catch, kinda gnarly.

oliverlloyd: > kinda gnarly

Yeah, I was hoping there'd be a nicer way to do this that didn't invovle messy logic and _re_setting state like this but this has to happen on the client and after page size selection so I think this is it

PR #148, opened 2020-03-30
TS declaration file

oliverlloyd:
## What does this change?
This fixes how we genreated the typescript declaration file

Props to @gtrufitt 

## Why?
We only need to generate the types for the App interface, not the entire component stack, and we want the file to be in a compatible format

https://trello.com/c/j4Tpddez/1379-ts-declaration

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -1,4 +1,10 @@
-export type Pillar = "news" | "opinion" | "sport" | "culture" | "lifestyle";
+export type Pillar =
+  | "news"
+  | "opinion"
+  | "sport"
+  | "culture"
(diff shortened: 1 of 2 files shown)


src/types.tsx
```diff
@@ -1,4 +1,10 @@
-export type Pillar = "news" | "opinion" | "sport" | "culture" | "lifestyle";
+export type Pillar =
+  | "news"
+  | "opinion"
+  | "sport"
+  | "culture"
+  | "lifestyle"
+  | "labs";
```
oliverlloyd: Little bit of boyscouting here (and also to ensure things work okay)

PR #152, opened 2020-03-31
Type safety to fixtures

liywjl:
## What does this change?
Move fixtures into `src/`

## Why?
Add type safety to fixtures

### DIFF
--- a/src/fixtures/comment.ts
+++ b/src/fixtures/comment.ts
@@ -1,4 +1,6 @@
-export const comment = {
+import { CommentType } from "../types";
+
+export const comment: CommentType = {
   id: 138809272,
   body:
     "<p>Begone ye self-serving Tory isolationists.</p> <p>Never has there been a more significant time for Big Government and public health provision. Nature has spoken.</p>",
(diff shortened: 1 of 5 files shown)


src/lib/mockFetchCalls.js
```diff
@@ -1,7 +1,7 @@
 import fetchMock from "fetch-mock";
 
-import { discussion } from "../../fixtures/discussion";
-import { comment } from "../../fixtures/comment";
+import { discussion } from "../fixtures/discussion.ts";
```
oliverlloyd: Is the .ts needed?

src/fixtures/comment.ts
```diff
@@ -50,7 +41,7 @@ export const comment = {
           "https://discussion.guardianapis.com/discussion-api/comment/138809272",
         isoDateTime: "2020-03-09T19:09:28Z",
         date: "09 March 2020 7:09pm",
-        commentId: "138809272",
+        commentId: 138809272,
```
oliverlloyd: Is this part of the move or the start of something else?

src/types.tsx
```diff
@@ -119,7 +119,7 @@ export interface UserProfile {
 
 export interface DiscussionResponse {
   status: string;
-  page: number;
+  currentPage: number;
```
oliverlloyd: Is this related to moving the file?

liywjl: I assume the fixtures come from actual API calls. If so there does not exist a `page` key only `currentPage`, so i made the relavant changes

src/fixtures/comment.ts
liywjl: for the sake of keeping this PR small im moving everything to number, in the next PR I will move them to all strings (right not they error types)

src/lib/mockFetchCalls.js
liywjl: good catch

PR #154, opened 2020-03-31
Refactor placeholder styling

oliverlloyd:
## What does this change?
This fixes a problem where the placeholder font was coloured incorrectly cross browser. It also refacotrs how we set placeholder foprmatting in general.

There are 3 states the placeholder text can be in:

### Reader is replying, form open
Screenshot 2020-03-31 at 11 52 51

### Reader is not replying, form open
Screenshot 2020-03-31 at 11 50 33

### Reader is not replying, form closed
Screenshot 2020-03-31 at 11 50 23

## Why?
To communicate well to the user what the form is for

https://trello.com/c/ZekHLd7m/1367-chrome-placeholder-colour

### DIFF
--- a/src/components/CommentForm/CommentForm.tsx
+++ b/src/components/CommentForm/CommentForm.tsx
@@ -65,21 +65,25 @@ const greyPlaceholder = css`
 const blackPlaceholder = css`
   ::-webkit-input-placeholder {
     font-weight: bold;
+    opacity: 1;
     color: ${neutral[0]};
   }
 
   ::-moz-placeholder {
     font-weight: bold;
+    opacity: 1;
     color: ${neutral[0]};
   }
 
   ::-ms-placeholder {
     font-weight: bold;
+    opacity: 1;
     color: ${neutral[0]};
   }
 
   ::placeholder {
     font-weight: bold;
+    opacity: 1;
     color: ${neutral[0]};
   }
 `;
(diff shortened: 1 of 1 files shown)


src/components/CommentForm/CommentForm.tsx
```diff
@@ -43,15 +43,48 @@ const commentTextArea = css`
   }
 `;
 
-const placeholderCommentStyles = css`
+// https://stackoverflow.com/questions/9451902/changing-an-inputs-html5-placeholder-color-with-css-does-not-work-on-chrome
+const greyPlaceholder = css`
+  ::-webkit-input-placeholder {
```
liywjl: I thought emotion would compile down our CSS using Babel to support all browsers, so we can just simply use `::placeholder`?

https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin-emotion

worth double checking what CSS is used on rendering

src/components/CommentForm/CommentForm.tsx
```diff
@@ -43,15 +43,48 @@ const commentTextArea = css`
...
   ::placeholder {
-    font-weight: bold;
-    color: black;
+    color: ${neutral[46]};
   }
 `;
-const placeholderReplyStyles = css`
+
+const blackPlaceholder = css`
+  ::-webkit-input-placeholder {
```
liywjl: same for here

src/components/CommentForm/CommentForm.tsx
oliverlloyd: You're completely right! The problem is solved with the opacity trick and these prefixes aren't needed.

PR #156, opened 2020-03-31
Add Flex component

liywjl:
## What does this change?
Add Flex component

## Why?
To avoid having to reimplement flexbox logic for Col and Row

### DIFF
--- a/src/components/Flex/Flex.tsx
+++ b/src/components/Flex/Flex.tsx
@@ -1,0 +1,29 @@
+import React from "react";
+import { css } from "emotion";
+
+type Props = {
+  children: React.ReactNode;
+  direction?: "row" | "column";
+  justify?: "flex-start" | "space-between"; // Extend as required
+  fullWidth?: boolean;
+};
+
+export const Flex = ({
+  children,
+  direction = "row",
+  justify = "flex-start",
+  fullWidth = false
+}: Props) => (
+  <div
+    className={css`
+      display: flex;
+      flex-direction: ${direction};
+      justify-content: ${justify};
+      ${fullWidth} {
+        width: 100%;
+      }
+    `}
+  >
+    {children}
+  </div>
+);


src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,29 @@
...
+  direction?: "row" | "column";
+  justify?: "flex-start" | "space-between"; // Extend as required
+  fullWidth?: boolean;
+};
+
+export const Flex = ({
+  children,
+  direction = "row",
+  justify = "flex-start",
+  fullWidth = false
```
oliverlloyd: My preference here would be to not try to contain all flex styling into one abstraction. If you're going to build a library of abstractions it's better to keep them declarative and readable by using smaller components like `Row` and `Column`.

This creates better boundaries and also guards against `Flex` becomming a bucket for all styles that involve `display: flex` which is a lot

src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { css } from "emotion";
+
+type Props = {
+  children: React.ReactNode;
+  direction?: "row" | "column";
+  justify?: "flex-start" | "space-between"; // Extend as required
+  fullWidth?: boolean;
```
oliverlloyd: This doesn't really relate to `Flex` and makes it more of a generic container which runs the risk of stretching the definition for this abstraction.

liywjl: `fullWidth` is a very common thing I use with flex, as we agreed that we should not use classNames, I feel it makes sense to add this here.

without it, this component would become pointless, as you would have no way of setting full width on flex components.

src/components/Flex/Flex.tsx
oliverlloyd: Also, there are five instances where we already have `<Row>` declared in the code and four with `<Column.`. This strongly suggests there's a use case for abstracting that pattern

liywjl: The logic of both components are near identical, having them as one component would avoid us from having to update 2 components when we add a new prop (but granted it is likely we wont be updating this regularly).

Also this is already implemented in DCR.

I feel this is more just a question of taste. So @gtrufitt would you have a preference?

gtrufitt: My preference would be maybe to publish the component from DCR and import it into discussion rendering but that's probably not what you wanna here :D. Any case, consistency between the two is probably the most useful.

oliverlloyd: Which is more readable 🤷‍♂ 

```
            <Flex direction="column">
              <Flex direction="row">
                <div>Thing</div>
              </Flex>
              <Flex direction="row">
                <Flex direction="column">
                  <div>Thing</div>
                  <div>Thing</div>
                  <div>Thing</div>
                </Flex>
              </Flex>
            </Flex>
```

```

            <Column>
              <Row>
                <div>Thing</div>
              </Row>
              <Row>
                <Column>
                  <div>Thing</div>
                  <div>Thing</div>
                  <div>Thing</div>
                </Column>
              </Row>
            </Column>
```

liywjl: you have convinced me, `<Row>` / `<Column>` it is

src/components/Flex/Flex.tsx
liywjl: do you have a suggestion for the `fullWidth` work around? or are we ok to keep this?

src/components/Flex/Flex.tsx
oliverlloyd: If we do implement this new Flex to shorthandfor  rows and columns then we should definately not do it and then leave the other pattern in the codebase as well. If we do  replace 'Row' and 'Column' then I'd much rather have it done properly rather than some places doing it one way and others doing it another

liywjl: we use Flex & Row/Column are in DCR . I feel like this was done locally just to help readability, but as a result it wasn't done with a lot of for-sight, we should just agree on a common abstraction, regardless of it is one or the other

liywjl:
based on discussion we will be going with a different implementation https://github.com/guardian/discussion-rendering/pulls

PR #168, opened 2020-04-01
Reverse workflow names

oliverlloyd:
## What does this change?
Swaps the workflow and job names around

## Why?
because the workflow name is often used by itself in emails and in other lists and emojis are cute and everything but text is more useful

https://trello.com/c/QN9W8Rh1/1386-reverse-workflow-names


oliverlloyd:
Well, that didn't work

PR #169, opened 2020-04-01
Refactor comment css variables

liywjl:
## What does this change?
Rename and extract flex Row/Column from css consts in Comment component

## Why?
Comment will need clearer definition of Row and Column elements to help build mobile layouts

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -46,11 +46,6 @@ const commentControlsButton = (pillar: Pillar) => css`
   }
 `;
 
-const spaceBetween = css`
-  display: flex;
-  justify-content: space-between;
-`;
-
 const commentCss = css`
   display: block;
   clear: left;
@@ -94,7 +89,7 @@ const selectedStyles = css`
   padding-right: ${space[2]}px;
 `;
 
-const avatarMargin = css`
+const rightMarginStyles = css`
   margin-right: ${space[2]}px;
 `;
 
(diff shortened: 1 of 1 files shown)


src/components/Comment/Comment.tsx
```diff
@@ -291,7 +283,13 @@ export const Comment = ({
                 ) : (
                   <></>
                 )}
-                <div className={timestampWrapperStyles}>
+                <div
+                  className={cx(
+                    timestampWrapperStyles,
+                    marginLeftStyles,
+                    flexColStyles
```
oliverlloyd, line 290: All of these abstractions are only ever used here so we can simplify this a lot by just having `timstampWrapperStyles` - there's no need for the other variables

src/components/Comment/Comment.tsx
```diff
@@ -154,35 +146,35 @@ const linkStyles = (pillar: Pillar) => css`
...
 const Row = ({ children }: { children: JSX.Element | JSX.Element[] }) => (
-  <div
-    className={css`
-      display: flex;
-      flex-direction: row;
-    `}
-  >
-    {children}
-  </div>
+  <div className={flexRowStyles}>{children}</div>
```
oliverlloyd, line 177: We already have Row and Column defined in lots of other places in the codebase and we intend to abstract that into its own component so we should expect to use that here.

Not a blocker, we can define the css outside as well, but when we abstract we should expect this to be replaced

PR #170, opened 2020-04-01
Optimistic replies v3

liywjl:
## What does this change?
Add optimistic reply rendering

## Why?
We have this logic already for comments, for consistency this logic should apply to replies

https://trello.com/c/pBVxrFs9/1340-optimistic-replies


.babelrc.js
```diff
@@ -4,5 +4,10 @@ module.exports = {
     "@babel/preset-typescript",
     "@babel/preset-react",
     "@emotion/babel-preset-css-prop"
-  ]
+  ],
+  env: {
+    test: {
+      plugins: ["@babel/plugin-transform-runtime"]
```
liywjl, line 10: this was used to fix a runtime issue in the unit tests

PR #172, opened 2020-04-02
Mute other users

oliverlloyd:
## What does this change?
This adds a feature to allow readers to mute other user so their posts no longer show for them on that particular device.

## Why?
This was one of the requests made during the meeting with the moderators

https://trello.com/c/pELxIx7Z/1279-wishlist-mute-a-user


src/components/Comment/Comment.tsx
```diff
@@ -384,12 +420,30 @@ export const Comment = ({
                       </button>
                     )}
                 </div>
-                <div>
+                <Row>
+                  {/* You can't mute unless logged in and you can't yourself */}
+                  {user && comment.userProfile.userId !== user.userId ? (
+                    <div className={buttonHeightOverrides}>
```
liywjl, line 426: the variations between this and `AbuseReportForm` cause the exact position to be slightly off, and for the on highlight to be different.

Screenshot 2020-04-02 at 13 09 23
Screenshot 2020-04-02 at 13 09 28

This should be fixed in another PR by changing `AbuseReportForm` to apply src foundations as it is due to src-foundations

oliverlloyd: Good spot. Agreed, lets move the abuse form over to src as well, and maybe do the same for Reply, Pick and Share at the same time?

liywjl: yes agreed!

oliverlloyd:
> `writeMutes` is not used

Fixed 👍

PR #178, opened 2020-04-02
Remove hex references

liywjl:
## What does this change?
Remove hard coded Hex and replace with src-foundations

## Why?
Allows css to have correct theming. Consistency with the rest of the repo

https://trello.com/c/AnSZs6Bu/1376-remove-hex-colours

### DIFF
--- a/src/components/Badges/Badges.tsx
+++ b/src/components/Badges/Badges.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { css, cx } from "emotion";
 
-import { space } from "@guardian/src-foundations";
+import { space, neutral, brand } from "@guardian/src-foundations";
 import { textSans } from "@guardian/src-foundations/typography";
 
 const staffBadge = css`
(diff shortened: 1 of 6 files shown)


src/components/RecommendationCount/RecommendationCount.tsx
```diff
@@ -2,14 +2,17 @@ import React, { useState } from "react";
...
 import { palette } from "@guardian/src-foundations";
 
 import { recommend } from "../../lib/api";
+import { Pillar } from "../../types";
 
 type Props = {
   commentId: number;
   initialCount: number;
   alreadyRecommended: boolean;
+  pillar: Pillar;
```
oliverlloyd, line 15: Recommendation doesn't use pillar for background colour, it's always this blue. I think you can grab it from the `brand` palette

Screenshot 2020-04-02 at 18 25 22

src/components/Badges/Badges.tsx
```diff
@@ -24,14 +24,14 @@ const labelText = css`
 `;
 
 const staffIcon = css`
-  fill: #005689;
+  fill: ${brand[500]};
```
oliverlloyd, line 27: I don't think 500 is right here, it seems too bright

PR #182, opened 2020-04-03
Fix button spacing

liywjl:
## What does this change?
Wrap link with button styles and remove padding-top from buttons and links

## Why?
Buttons did not have correct spacing between each other, this was due to the fact that some buttons were implicitly links and did not implement wrapper styles. In order to keep buttons and links to the same margin we removed top padding to all of them.

https://trello.com/c/rgRHUKEV/1394-discussion-button-spacing

## Before
Screenshot 2020-04-03 at 10 42 41

## After
Screenshot 2020-04-03 at 10 42 53

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -50,7 +50,13 @@ const commentControlsButton = (pillar: Pillar) => css`
   :hover {
     text-decoration: underline
   }
-  padding-top: 0px; /* In order to remove variations between buttons and a tags */
+  padding-top: 0px; /* In order to remove variations between <button> and <a> tags */
+`;
+
+const commentControlsLink = (pillar: Pillar) => css`
+  ${textSans.xsmall({ fontWeight: "bold" })}
+  margin-right: ${space[2]}px;
+  color: ${palette[pillar][400]};
 `;
 
 const spaceBetween = css`
@@ -226,6 +232,7 @@ export const Comment = ({
   toggleMuteStatus
 }: Props) => {
   const commentControlsButtonStyles = commentControlsButton(pillar);
+  const commentControlsLinkStyles = commentControlsLink(pillar);
   const [isHighlighted, setIsHighlighted] = useState<boolean>(
     comment.isHighlighted
   );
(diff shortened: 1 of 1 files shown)


src/components/Comment/Comment.tsx
```diff
@@ -507,24 +508,28 @@ export const Comment = ({
...
+                          Reply
                         </button>
                       ) : (
-                        <Row>
+                        <div
+                          className={cx(
+                            commentControlsButtonStyles,
+                            flexRowStyles
+                          )}
+                        >
```
oliverlloyd, line 524: This could be a Button. Then we can pass the svg in as a prop. We actually already have this reply button in the code here

https://github.com/guardian/discussion-rendering/blob/577c602da6a650a4605e53d6772e450e35229fdb/src/components/Comment/Comment.tsx#L370

src/components/Comment/Comment.tsx
```diff
@@ -507,24 +499,30 @@ export const Comment = ({
...
-                          </div>
+                          <ReplyArrow />
+                          Reply
                         </button>
                       ) : (
-                        <Row>
-                          <ReplyArrow />
-                          <a
-                            className={linkStyles(pillar)}
+                        <div className={commentControlsButtonStyles}>
```
oliverlloyd, line 510: `commentControlsButtonStyles` seems to being applied twice here. And actually, now that we're using link, a lot of those styles are redundant

src/components/Comment/Comment.tsx
```diff
@@ -507,24 +499,30 @@ export const Comment = ({
                           onClick={() => setCommentBeingRepliedTo(comment)}
                           className={cx(
                             commentControlsButtonStyles,
-                            removePaddingLeft
+                            removePaddingLeft,
+                            flexRowStyles
```
oliverlloyd, line 503: We wouldn't need these if we use the Link component from source foundation?

liywjl: this is a button HTML element, so I am unsure what you mean. Nonetheless it does raise the point that we should convert the buttons here to src foundations. I prefer to do it in another PR

PR #183, opened 2020-04-03
Abstracted Row & Column

oliverlloyd:
## What does this change?
This abstracts the `Row` and `Column` components that were being used in several places in the codebase up into their own component.

## Why?
For speed of development

https://trello.com/c/BJslsCLx/1396-abstract-row-abstract-row-column

### DIFF
--- a/src/components/Column/Column.tsx
+++ b/src/components/Column/Column.tsx
@@ -1,0 +1,15 @@
+import React from "react";
+import { css } from "emotion";
+
+type Props = { children: JSX.Element | JSX.Element[] };
+
+export const Column = ({ children }: Props) => (
+  <div
+    className={css`
+      display: flex;
+      flex-direction: column;
+    `}
+  >
+    {children}
+  </div>
+);
--- a/src/components/LoadingPicks/LoadingPicks.tsx
+++ b/src/components/LoadingPicks/LoadingPicks.tsx
@@ -4,6 +4,8 @@ import { css, keyframes } from "emotion";
 import { space } from "@guardian/src-foundations";
 import { neutral } from "@guardian/src-foundations/palette";
 
+import { Row } from "../Row/Row";
+
 const BACKGROUND_COLOUR = neutral[93];
 
 const shimmer = keyframes`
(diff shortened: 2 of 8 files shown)


src/components/Column/Column.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { css } from "emotion";
+
+type Props = { children: JSX.Element | JSX.Element[] };
+
+export const Column = ({ children }: Props) => (
+  <div
+    className={css`
+      display: flex;
```
liywjl, line 9: by having inlines styles we generate a new key for every use. In order for them to share the same key and reduce the css bundle size we could maybe define a const outside the component?

src/components/Row/Row.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { css } from "emotion";
+
+type Props = { children: JSX.Element | JSX.Element[] };
+
+export const Row = ({ children }: Props) => (
+  <div
+    className={css`
+      display: flex;
```
liywjl, line 9: same thing here regarding inline styles

src/components/LoadingPicks/LoadingPicks.tsx
```diff
@@ -62,7 +64,11 @@ const pickBoxStyles = css`
   ${shimmerStyles}
 `;
 
-const Column = ({ children }: { children: JSX.Element | JSX.Element[] }) => (
+const FullWidthColumn = ({
```
liywjl, line 67: 😅 there was nothing full width here, good catch

src/components/Column/Column.tsx
oliverlloyd: > by having inlines styles we generate a new key for every use

I never realised this. Is there a link to the docs about this? We use the inline pattern a lot in the codebase, I wonder if there's a solution higher up that we could look at rather than having to refactor everything?

gtrufitt: Yeah interesting!

liywjl: I ran a couple tests to see how keys are generated. 

They keys seem generated by component scope, so you are right @oliverlloyd  there is actually no difference. 

I assumed css command worked like a function and was concerned that calling it inline would be like calling `css(...)` inline causing a recalculation every time.

Some findings I made, if you define
```
const someStyles = css`
  width: 30px;
  color: blue;
`

css`
  width: 30px;
  color: blue;
`
 ```
both will generate different CSS keys, so you either need to stick to one or the other in the same Component.

oliverlloyd: Awesome investigation! Shame there's not a perf win to be had but it's useful to learn about the inline vs. const difference inside a component 👍

PR #184, opened 2020-04-03
Fix mute and report alignment

liywjl:
## What does this change?
Use src foundations for both Report and Mute button, add the same wrapper styles for each

## Why?
The implementation of each button was different, causing alignment issues. By implementing both buttons using src-foundations styles, and giving them the same wrapper styles we are able to fix the alignment issue 

https://trello.com/c/cXEmgSuX/1395-fix-report-and-mute-button-alignment

## Before
Screenshot 2020-04-03 at 10 47 05

## After
Screenshot 2020-04-03 at 11 08 16


liywjl:
broken down into smaller PRs

PR #188, opened 2020-04-03
Use Row & Column

oliverlloyd:
## What does this change?
This makes use of the new `Row` and `Column` componments

## Why?
To make the code more readable

https://trello.com/c/8BgiQ9vX/1397-use-row-column

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -369,66 +369,29 @@ export const App = ({
 
   return (
     <Column>
-      {user && !isClosedForComments && (
-        <CommentForm
-          shortUrl={shortUrl}
-          onAddComment={onAddComment}
-          user={user}
-        />
-      )}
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -372,67 +368,30 @@ export const App = ({
...
-      <Filters
-        filters={filters}
-        onFilterChange={onFilterChange}
-        totalPages={totalPages}
-        commentCount={commentCount}
-      />
-      {showPagination && (
-        <Pagination
+    <Column>
+      <>
```
oliverlloyd, line 372: This wrapper fragment is needed to prevent typescript complaining that we're pass 'undefined' or 'false' as children

liywjl: I much prefer wrapping in fragments , its easier on the eyes 👍

PR #189, opened 2020-04-03
Center recommendation count button

liywjl:
## What does this change?
- Match SVG width to content's width
- change margin left styles in wrapper div to reflect content change

## Why?
The SVG wasn't being centered correctly

## Before
Screenshot 2020-04-03 at 13 12 49

## After
Screenshot 2020-04-03 at 13 04 11


oliverlloyd:
Interesting how this was much worse in Chrome vs. Firefox. You solution works great for both though 👍

PR #196, opened 2020-04-06
Comment buttons to src foundation buttons

liywjl:
## What does this change?
Migrate comment buttons to src foundations button rather than default HTML button

## Why?
We should use src foundations by default rather than HTML buttons

https://trello.com/c/hyAVeTOF/1398-bug-comment-box-still-displayed-when-comments-are-closed


src/components/Comment/Comment.tsx
```diff
@@ -488,17 +496,22 @@ export const Comment = ({
...
-                        </button>
+                          <Button
+                            priority="tertiary"
+                            size="small"
+                            onClick={() => setCommentBeingRepliedTo(comment)}
+                          >
+                            <ReplyArrow />
+                            Reply
+                          </Button>
+                        </div>
```
oliverlloyd, line 514: Button has the `icon` and `iconPosition` props we can use here

src/components/Comment/Comment.tsx
```diff
@@ -40,17 +40,21 @@ const commentControls = css`
...
+const commentControlsButtonWrapper = (pillar: Pillar) => css`
+  button {
+    ${textSans.xsmall({ fontWeight: "bold" })}
+    margin-right: ${space[2]}px;
+    color: ${palette[pillar][400]};
+    padding-left: ${space[2]}px;
+    padding-right: ${space[2]}px;
+    background-color: transparent;
+    border: 0;
+    cursor: pointer;
```
oliverlloyd, line 52: Is this needed? I'm guessing Button does this already

src/components/Comment/Comment.tsx
```diff
@@ -40,17 +40,21 @@ const commentControls = css`
...
-  color: ${palette[pillar][400]};
-  background-color: transparent;
-  border: 0;
-  cursor: pointer;
-  :hover {
-    text-decoration: underline
+const commentControlsButtonWrapper = (pillar: Pillar) => css`
+  button {
+    ${textSans.xsmall({ fontWeight: "bold" })}
+    margin-right: ${space[2]}px;
```
oliverlloyd, line 46: We probably shouldn't be applying margins to elements like this. The spacing around an element should, ideally, be context specific

src/components/Comment/Comment.tsx
```diff
@@ -173,8 +177,10 @@ const flexRowStyles = css`
   flex-direction: row;
 `;
 
-const removePaddingLeft = css`
-  padding-left: 0px;
+const removeButtonPaddingLeft = css`
+  button {
+    padding-left: 0px;
+  }
```
oliverlloyd, line 183: This rule conflicts with how padding is set in the `commentControlsButtonWrapper ` css above

oliverlloyd:
@SiAdcock  Hmm, looks like there's already [something called](http://theguardian.design/2a1e5182b/p/435225--button/b/86f344) `LinkButton` but that' an a tag styled to look like a `Button`. What we need is a button tag styled to look like a `Link`

src/components/Comment/Comment.tsx
liywjl: I tired this, but it looks like a bunch of unwanted styling (colour and padding) applied to the SVG, we can override them as-well knowing we are doing something similar to the Button styles

Screenshot 2020-04-06 at 13 42 44

src/components/Comment/Comment.tsx
liywjl: this raises an interesting issue, we need to remove padding left from only the 1st button and that button is not necessarily the reply button.

src/components/Comment/Comment.tsx
oliverlloyd: It's because it's expecting an icon of a certain size. If you used one of the Source icons it works fine. Check out `svgOverrides`

src/components/Comment/Comment.tsx
liywjl: thx for catching

src/components/Comment/Comment.tsx
liywjl: yeah it looks like this was unnecessary, good catch

src/components/Comment/Comment.tsx
oliverlloyd: I think a good pattern in these cases is to make the components margin/padding agnostic and to extract spacing into context specific components. It also solves the only pad when not first item problem.

An example of this is in `FirstCommentWelcome`:

```
        <Row>
          <Button size="small" onClick={() => submitForm(userName)}>
            Post your comment
          </Button>
          <div
            className={css`
              width: ${space[3]}px;
            `}
          ></div>
          <Button size="small" priority="tertiary" onClick={cancelSubmit}>
            Cancel
          </Button>
        </Row>
```

oliverlloyd: Like this, if we extracted the Button logic into a separate component it would read as:

```
        <Row>
          <ButtonLink size="small" onClick={() => doStuff()}>
            Reply
          </ButtonLink>
          <div
            className={css`
              width: ${space[3]}px;
            `}
          ></div>
          <ButtonLink size="small"onClick={otherStuff()}>
            Share
          </ButtonLink>
        </Row>
```

You could maybe (🤷‍♂) also abstract the spacing css to give:

```
        <Row>
          <ButtonLink size="small" onClick={() => doStuff()}>
            Reply
          </ButtonLink>
          <Space amount={3} />
          <ButtonLink size="small"onClick={otherStuff()}>
            Share
          </ButtonLink>
        </Row>
```

src/components/Comment/Comment.tsx
```diff
@@ -219,6 +217,14 @@ const ReplyArrow = () => (
   </svg>
 );
 
+const Space = ({ amount }: { amount: 1 | 2 | 3 | 4 | 5 | 6 | 9 | 12 | 24 }) => (
+  <div
+    className={css`
+      padding-right: ${space[amount]}px;
```
oliverlloyd, line 223: I think you're better off using width here, just from a semantic point of view

SiAdcock:
@oliverlloyd would the [`priority="tertiary"`](https://www.theguardian.design/2a1e5182b/p/435225--button/b/86f344) button serve for your purpose?

src/components/Comment/Comment.tsx
```diff
@@ -231,7 +237,9 @@ export const Comment = ({
   isMuted,
   toggleMuteStatus
 }: Props) => {
-  const commentControlsButtonStyles = commentControlsButton(pillar);
+  const commentControlsButtonWrapperStyles = commentControlsButtonWrapper(
+    pillar
+  );
```
oliverlloyd, line 242: Would it make things simpler to rename this to `buttonOverrides(pillar)` and use the function directly?

src/components/Comment/Comment.tsx
```diff
@@ -40,17 +40,19 @@ const commentControls = css`
...
-  :hover {
-    text-decoration: underline
+const commentControlsButtonWrapper = (pillar: Pillar) => css`
+  button {
+    ${textSans.xsmall({ fontWeight: "bold" })}
+    color: ${palette[pillar][400]};
+    background-color: transparent;
+    border: 0;
+    :hover {
+      text-decoration: underline
```
oliverlloyd, line 50: You will probably also need to override `text-decoration-color`

src/components/Comment/Comment.tsx
```diff
@@ -40,17 +40,19 @@ const commentControls = css`
...
-  border: 0;
-  cursor: pointer;
-  :hover {
-    text-decoration: underline
+const commentControlsButtonWrapper = (pillar: Pillar) => css`
+  button {
+    ${textSans.xsmall({ fontWeight: "bold" })}
+    color: ${palette[pillar][400]};
+    background-color: transparent;
+    border: 0;
```
oliverlloyd, line 48: Do we need this if we're using `tertiary`?

oliverlloyd:
> would the [`priority="tertiary"`](https://www.theguardian.design/2a1e5182b/p/435225--button/b/86f344) button serve for your purpose?

We're using that already but it has a number of issues:

It's rather wide and tall, pushing content away from it in a way that a 'link' wouldn't normally
It has a `white` background that doesn't play well on top of grey backgrounds
The font colour is blue, as is the text underline
The font size is larger than our default if `textSans.xsmall`

We're obviously customising a lot here but I still think there's benefit in using Source as the base.

src/components/Comment/Comment.tsx
```diff
@@ -40,17 +40,19 @@ const commentControls = css`
...
+    :hover {
+      text-decoration: underline
+    }
+    /* Button height overrides */
+    height: 18px;
+    min-height: 18px;
```
oliverlloyd, line 54: This css is already defined in the same file as:

```
const buttonHeightOverrides = css`
  button {
    height: 18px;
    min-height: 18px;
  }
`;
```

oliverlloyd:
Would it make sense to abstract this work into a `ButtonLink` component? I notice we already have:

```
                    <div className={buttonHeightOverrides}>
                      <Button
                        priority="tertiary"
                        size="small"
                        onClick={() =>
                          toggleMuteStatus(comment.userProfile.userId)
                        }
                      >
                        <span className={muteReportTextStyles}>Mute</span>
                      </Button>
                    </div>
```

For the mute and report buttons where a lot of this work has been done already

SiAdcock:
@oliverlloyd makes a lot of sense, thanks for the explanation. Yes, I think this would be a useful addition. Perhaps it could be an additional export of `@guardian/src-link`... from a design perspective, and from (sighted) users' perspectives, it looks like a link and will be treated as such. When this is merged, I'll raise an issue against Source to consider promoting this code, assuming it serves you well 🙂

src/components/Comment/Comment.tsx
liywjl: weirdly it seems to work without, however this led me down a rabbit whole in which i found that the links didnt have the same `text-decoration-color`

liywjl: I am going to apply `text-decoration-color` here nonetheless incase there is variation between browsers

PR #204, opened 2020-04-08
Use PillarButton

oliverlloyd:
## What does this change?
Makes use of the PillarButton in place of various custom overrides. To do this we are now passing `pillar` as a prop to those places where it was needed

## Why?
Becuase we want consistent button styles that support pillar colours

https://trello.com/c/vTtzthKP/1407-view-more-comments-not-applying-pillar-styles

### DIFF
--- a/src/components/CommentContainer/CommentContainer.tsx
+++ b/src/components/CommentContainer/CommentContainer.tsx
@@ -1,220 +1,237 @@
-import React, { useState, useEffect } from "react";
-import { css, cx } from "emotion";
+import React, { useState, useEffect } from 'react';
+import { css, cx } from 'emotion';
 
-import { space, neutral, palette, border } from "@guardian/src-foundations";
+import { space, neutral, palette, border } from '@guardian/src-foundations';
 
-import { Pillar, CommentType, UserProfile, ThreadsType } from "../../types";
-import { CommentForm } from "../CommentForm/CommentForm";
-import { Comment } from "../Comment/Comment";
-import { Row } from "../Row/Row";
-import { CommentReplyPreview } from "../CommentReplyPreview/CommentReplyPreview";
+import { Pillar, CommentType, UserProfile, ThreadsType } from '../../types';
+import { CommentForm } from '../CommentForm/CommentForm';
+import { Comment } from '../Comment/Comment';
+import { Row } from '../Row/Row';
+import { CommentReplyPreview } from '../CommentReplyPreview/CommentReplyPreview';
 
-import { getMoreResponses } from "../../lib/api";
+import { getMoreResponses } from '../../lib/api';
 
 type Props = {
-  baseUrl: string;
-  comment: CommentType;
-  pillar: Pillar;
-  isClosedForComments: boolean;
-  shortUrl: string;
-  user?: UserProfile;
-  threads: ThreadsType;
-  commentBeingRepliedTo?: CommentType;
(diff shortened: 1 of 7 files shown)


src/components/CommentForm/CommentForm.tsx
```diff
@@ -426,39 +421,37 @@ export const CommentForm = ({
...
                 <div className={bottomContainer}>
-                    <div className={buttonContainerStyles}>
-                        <Button
-                            type="submit"
-                            size="small"
-                            data-testid="post-comment"
-                        >
-                            <div className={buttonTextStyles}>
+                    <Row>
+                        <>
```
liywjl, line 425: unnecessary

src/components/CommentForm/CommentForm.tsx
```diff
@@ -359,13 +360,7 @@ export const CommentForm = ({
...
     return (
         <>
-            <form
-                className={formWrapper}
-                onSubmit={e => {
-                    e.preventDefault();
-                    submitForm();
-                }}
-            >
+            <form className={formWrapper}>
```
liywjl, line 363: why was `onSubmit` logic removed?

src/components/CommentContainer/CommentContainer.tsx
```diff
@@ -1,220 +1,237 @@
...
+    toggleMuteStatus: (userId: string) => void;
 };
 
 const nestingStyles = css`
-  list-style-type: none;
-  padding-left: ${space[2]}px;
-  margin-left: ${space[12] + "px"};
+    list-style-type: none;
+    padding-left: ${space[2]}px;
+    margin-left: ${space[12] + 'px'};
```
liywjl, line 32:
```suggestion
    margin-left: ${space[12]}px;
```

src/components/CommentForm/CommentForm.tsx
oliverlloyd: It's needed because Row expects an element, not an element or string

src/components/CommentForm/CommentForm.tsx
oliverlloyd: Because we can't pass `type='submit'` to `Button`. Unless I'm wrong about that?

src/components/CommentForm/CommentForm.tsx
liywjl: we could use `React.ReactNode` rather  than `JSX.Element | JSX.Element[]` in Row then to avoid having to do this.

I found a great answer on these variations here: https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

liywjl: I wouldn't consider this essential for the PR through, so feel free to ignore

src/components/CommentForm/CommentForm.tsx
liywjl: You can:
```
const Button = ({
	priority,
	size,
	icon: iconSvg,
	iconSide,
	cssOverrides,
	children,
	...props
}: ButtonProps) =>
```
it will be extracted and applied using `...props`

src/components/CommentForm/CommentForm.tsx
oliverlloyd: Nice. That's interesting. I'll look into this separately

src/components/CommentForm/CommentForm.tsx
oliverlloyd: Done

PR #208, opened 2020-04-09
Use pillar to style dropdown check

oliverlloyd:
## What does this change?
Changes the colour of the checkmark used for selected Dropdown items based on `pillar`

Screenshot 2020-04-09 at 16 38 23

## Why?
To enhance the pillar styling of discussion

https://trello.com/c/IQ0WaZDC/1429-style-dropdown-by-pillar

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -386,6 +386,7 @@ export const App = ({
                     />
                 )}
                 <Filters
+                    pillar={pillar}
                     filters={filters}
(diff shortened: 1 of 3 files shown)


oliverlloyd:
> Would be great to have a story for the drop down, shame we dont get any visual diffs

Looking at the code here we'd need to refactor things to extract the options list but I'm not convinced the test benefits justify the creation of a new component for showing list options. It's a lot of added complexity.

PR #209, opened 2020-04-09
No self aggrandissement

oliverlloyd:
## What does this change?
You can't recommend yourself

## Why?
These requests would fail quickly with api errors but it's better to not let the user make the request in the first place

https://trello.com/c/dPRzUgzV/1428-you-cant-recommend-yourself

### DIFF
--- a/src/components/TopPicks/TopPicks.tsx
+++ b/src/components/TopPicks/TopPicks.tsx
@@ -73,7 +73,7 @@ export const TopPicks = ({
                             comment={comment}
                             isSignedIn={isSignedIn}
                             userMadeComment={
-                                user &&
+                                !!user &&
                                 user.userId === comment.userProfile.userId
                             }
                         />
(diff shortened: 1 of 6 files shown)


src/components/TopPicks/TopPicks.tsx
```diff
@@ -46,14 +47,21 @@ const oneColCommentsStyles = css`
...
+    isSignedIn,
+}: Props) => {
     const leftColComments: CommentType[] = [];
     const rightColComments: CommentType[] = [];
     comments.forEach((comment, index) =>
         index % 2 === 0
             ? leftColComments.push(comment)
             : rightColComments.push(comment),
     );
+    console.log('user', user);
```
liywjl, line 64: console log

oliverlloyd, line 64:
```suggestion
```

PR #210, opened 2020-04-09
Scroll to filters

oliverlloyd:
## What does this change?
When a reader navigates to a new page, the scroll position of the window is set to the top of the comments list

## Why?
So the window position is reset to the first item in the list that the reader is navigating

https://trello.com/c/EZwWaRsV/1427-scroll-to-top-on-pagination

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -263,6 +263,7 @@ export const App = ({
     };
 
     const onPageChange = (page: number) => {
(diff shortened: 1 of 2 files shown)


oliverlloyd:
> Wouldn't this be solved by having some loading state being rendered?

We do show a loading state for the comments body but that doesn't scroll the page to the top of the view. The benefit to this change is it's easier to navigate from page to page because the control is snapped to the top.

PR #213, opened 2020-04-14
Upgrade emotion/core dependency

gtrufitt:
## What does this change?

Upgrades the `emotion/core` dependency to latest in `peerDependencies` (and removes from `devDependencies` as doesn't need to be in both). 

## Why?

Will upgrade in `dotcom-rendering` too, to dedupe dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,13 +24,13 @@
         "timeago.js": "^4.0.2"
     },
     "peerDependencies": {
-        "@emotion/core": "^10.0.27",
-        "emotion": "^10.0.27",
+        "@emotion/core": "^10.0.28",
         "@guardian/src-button": "^0.17.0",
         "@guardian/src-foundations": "^0.17.0",
         "@guardian/src-link": "^0.17.0",
         "@guardian/src-svgs": "^0.17.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,13 +24,13 @@
...
     "peerDependencies": {
-        "@emotion/core": "^10.0.27",
-        "emotion": "^10.0.27",
+        "@emotion/core": "^10.0.28",
         "@guardian/src-button": "^0.17.0",
         "@guardian/src-foundations": "^0.17.0",
         "@guardian/src-link": "^0.17.0",
         "@guardian/src-svgs": "^0.17.0",
         "@guardian/src-text-input": "^0.17.0",
+        "emotion": "^10.0.27",
```
oliverlloyd, line 33: Is there an `emotion@10.0.28`? Does it matter?

gtrufitt: Nope, there isn't, and I don't think it matters, they look like they're partly aligned but versioned independently https://github.com/emotion-js/emotion/blob/master/packages/core/CHANGELOG.md#patch-changes

PR #214, opened 2020-04-14
Use single tsconfig.json

gtrufitt:
## What does this change?

Removes the separate tsconfig files. @oliverlloyd I can't remember why this was done, I had a feeling that it was because the types were not working as expected within discussion rendering? This seems to not be the case but if you could check out and sanity check me, that'd be great.

## Why?

Less complexity.

Types in DCR and DIR.

Screen Shot 2020-04-14 at 16 24 32
Screen Shot 2020-04-14 at 16 24 16

And renders in DCR.

Screen Shot 2020-04-14 at 16 35 26


tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+    "compilerOptions": {
+        "target": "ES6",
+        "module": "esnext",
+        "lib": ["es2017", "es7", "es6", "dom", "DOM.Iterable"],
+        "outDir": "build",
```
gtrufitt, line 6: This is the new bit, that was possibly missing previously to spit `App.tsx` into the build folder.

gtrufitt:
> [This was the PR](https://github.com/guardian/discussion-rendering/pull/157) implementing this split. We were trying to solve the problem of the types not being built and published correctly. Looks like you've got around that 👍

Hopefully!

PR #218, opened 2020-04-15
Replace Reply Arrow SVG with src-svg indent component

liywjl:
## What does this change?
Use src-svg's component for the reply arrow

## Why?
always prefer src lib when possible

https://trello.com/c/5tuOL4gM/1426-use-sources-indent-svg-when-it-lands

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -6,6 +6,7 @@ import { from, until } from '@guardian/src-foundations/mq';
 import { neutral, border } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
 import { Link } from '@guardian/src-link';
+import { SvgIndent } from '@guardian/src-svgs';
 
 import { GuardianStaff, GuardianPick } from '../Badges/Badges';
 import { RecommendationCount } from '../RecommendationCount/RecommendationCount';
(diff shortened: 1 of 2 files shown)


src/components/Comment/Comment.tsx
```diff
@@ -144,6 +145,8 @@ const svgOverrides = css`
         fill: ${neutral[46]} !important;
         left: 3px !important;
         bottom: 0 !important;
+        width: 18px !important;
+        height: 18px !important;
```
oliverlloyd, line 149: If we still need to override the position and size of the icon then something is wrong. The point of using a Source icon was to remove the need for the overrides

src/components/CommentReplyPreview/CommentReplyPreview.tsx
```diff
@@ -23,7 +24,11 @@ const Space = ({ amount }: { amount: 1 | 2 | 3 | 4 | 5 | 6 | 9 | 12 | 24 }) => (
 );
 
 const fillGrey = css`
-    fill: ${neutral[46]};
+    svg {
+        fill: ${neutral[46]} !important;
+        width: 18px !important;
+        height: 18px !important;
```
oliverlloyd, line 30: _More_ overrides vs. less leaves us worse off than before

oliverlloyd:
Looks like we lost the icon on CommentReplyTo

liywjl:
need to wait for alignment SVG fix in @guardian/src-link v0.18.0

liywjl:
we can increment source foundations link later to fix alignment issues

PR #221, opened 2020-04-16
Add Initial OphanTrackingNames Type

gtrufitt:
## What does this change?

Adds the Ophan component and link names to the elements we want to track in discussion rendering.

## Why?

Existing tracking used to measure engagement + our requirement to measure the difference in engagement between DCR and Frontend.

## Types

Note: The inconsistencies in hyphenated, case and colon usage exists already so needs to be followed across for tracking reports. New tracking uses standardised lowercase, hyphenated format.

| Name *in code* | Done |
|-------------------------------|------|
| more-comments | ✅ | 
| Pagination view page 2 | ✅ | 
| Show more replies | ✅ |  
| Recommend comment | ✅ |  
| reply to comment | ✅ |  
| post comment | ✅ |  
| social-comment : facebook | 🚫 |  
| social-comment : twitter | 🚫 |  
| Open report abuse | ✅ |  
| Post report abuse | ✅ |  
| **NEW TRACKING** |  |  
| preview-comment | ✅ |  
| cancel-post-comment | ✅ |  
| formatting-controls-bold |  ✅ |  
| formatting-controls-italic |  ✅ |  
| formatting-controls-quote |  ✅ |  
| formatting-controls-link |  ✅ |  
| jump-to-comment-timestamp | ✅ |  
| commenter-profile-top-pick | ✅ |  
| mute-user | ✅|  
| unmute-user |✅ |  
| pick-comment |✅ |  
| unpick-comment |✅ |  
| reply-comment-hide |✅ | 
| reply-comment-show |✅ |  

https://trello.com/c/Xmg5bQcM

### DIFF
--- a/src/components/ButtonLink/ButtonLink.stories.tsx
+++ b/src/components/ButtonLink/ButtonLink.stories.tsx
@@ -24,7 +24,7 @@ export const Single = () => (
             alert('Clicked!');
         }}
         pillar="lifestyle"
-        linkName="ophanLink"
+        linkName=""
     >
         I'm a button but I look like a link. Click me
     </ButtonLink>
@@ -38,7 +38,7 @@ export const Small = () => (
         }}
         size="small"
         pillar="lifestyle"
-        linkName="ophanLink"
+        linkName=""
     >
         I'm small
     </ButtonLink>
@@ -47,15 +47,15 @@ Small.story = { name: 'a small button' };
 
 export const Group = () => (
     <Row>
-        <ButtonLink onClick={() => {}} pillar="culture" linkName="ophanLink">
+        <ButtonLink onClick={() => {}} pillar="culture" linkName="">
             Culture one
         </ButtonLink>
         <Space amount={3} />
-        <ButtonLink onClick={() => {}} pillar="news" linkName="ophanLink">
+        <ButtonLink onClick={() => {}} pillar="news" linkName="">
             News two
         </ButtonLink>
         <Space amount={3} />
-        <ButtonLink onClick={() => {}} pillar="sport" linkName="ophanLink">
+        <ButtonLink onClick={() => {}} pillar="sport" linkName="">
             Sport three
         </ButtonLink>
     </Row>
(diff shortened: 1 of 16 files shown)


src/components/ButtonLink/ButtonLink.stories.tsx
```diff
@@ -24,6 +24,7 @@ export const Single = () => (
             alert('Clicked!');
         }}
         pillar="lifestyle"
+        linkName="ophanLink"
```
jfsoul, line 27: pedantic, but lowercase hyphenate `ophanLink` if that's the casing we would prefer generally?

src/components/Comment/Comment.tsx
```diff
@@ -551,7 +553,9 @@ export const Comment = ({
                                                             iconSide="left"
                                                         >
                                                             {/* We use this span to scope the styling */}
-                                                            <span>Reply</span>
+                                                            <span data-link-name="reply to comment">
```
jfsoul, line 556: Not sure what frontend behaviour is, but should we differentiate this sign-in link from the above?

src/components/Pagination/Pagination.tsx
```diff
@@ -132,17 +133,20 @@ const Back = ({
...
-        className={chevronStyles(false)}
-        onClick={() =>
-            setCurrentPage(currentPage - 1 < 0 ? 0 : currentPage - 1)
-        }
-    >
-        <ChevronBack />
-    </button>
-);
+}) => {
+    const newPage = currentPage - 1 < 0 ? 0 : currentPage - 1;
```
jfsoul, line 137: Or `max(0, currentPage - 1)`?

src/components/ButtonLink/ButtonLink.stories.tsx
gtrufitt: Yeah, good point, these are actually placeholders so might just make them empty strings for `stories`

src/components/Comment/Comment.tsx
gtrufitt: That would be useful, but it's not differentiated on Frontend, so for now we probably need to keep it the same to avoid discrepancy in the date. When on full discussion-rendering on the platform we can probably separate.

src/components/Pagination/Pagination.tsx
gtrufitt: Nice!

PR #239, opened 2020-04-24
No comments

oliverlloyd:
## What does this change?
Adds some styling arond the message shown when there are no comments

## Why?
Technically, this message shouldn't be required but there's a bug in the api that means we sometimes end up in this state so some form of basic message is more graceful than nothing or a poorly styled message

https://trello.com/c/LG2f0Hfg/1519-todo-no-comments

### DIFF
--- a/src/App.stories.tsx
+++ b/src/App.stories.tsx
@@ -39,6 +39,7 @@ export const Default = () => (
                 'GU-Client': 'testClientHeader',
             }}
             expanded={false}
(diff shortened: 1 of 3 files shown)


oliverlloyd:
> Would be great to have a story for this?

This scenario should never happen so it's hard to test for, I'd need to break the code to force it or extract it out into a new component - neither are tempting, It's only there to manage a bug in the api so having a story would be confusing

PR #246, opened 2020-04-28
Restyle pagination

oliverlloyd:
## What does this change?
Refactors the styling used for the pagination buttons to prefer Source

Along the way, I move the margins from the left to the right so we are flush to the left now

### Before
Screenshot 2020-04-28 at 14 40 04

### After
Screenshot 2020-04-28 at 14 39 24

## Why?
There was an alignment issue with the icon in Firefox so rather than fix that I leant into Source but with overrides

### DIFF
--- a/src/components/Pagination/Pagination.stories.tsx
+++ b/src/components/Pagination/Pagination.stories.tsx
@@ -109,6 +109,20 @@ export const SevenPages = () => {
 };
 SevenPages.story = { name: 'with seven pages' };
 
+export const TwelvePages = () => {
+    const [page, setCurrentPage] = useState(1);
+    return (
+        <Pagination
+            totalPages={12}
+            currentPage={page}
+            setCurrentPage={setCurrentPage}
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@guardian/discussion-rendering",
     "description": "This codebase started as a hack day project by @gtrufitt and @nicl. The purpose is parity of the existing discussion application on Frontend using the discussion API (search for Private Repo).",
-    "version": "2.2.6",
```
gtrufitt, line 4: Isn't this on master?

oliverlloyd: Keen eyes! I forgot to merge the other PR. Done now and merged into here

paperboyo:
Aside: at least the rhs one doesn't look like it's [optically centred](https://medium.muz.li/optical-effects-9fca82b4cd9a) horizontally.

paperboyo:
Thanks, @oliverlloyd!

PR #249, opened 2020-05-01
Fix Report form

gtrufitt:
## What does this change?

- Prevents default on submission so that we don't page refresh
- Moves the API response error message so that it doesn't jolt the form
- Prevents going to API if the required field isn't complete
- Add success message

### DIFF
--- a/src/components/AbuseReportForm/AbuseReportForm.tsx
+++ b/src/components/AbuseReportForm/AbuseReportForm.tsx
@@ -176,7 +176,7 @@ export const AbuseReportForm: React.FC<{
                         Category
                     </label>
                     <select
-                        name="categoryId"
+                        name="category"
                         id="category"
                         onChange={(e: React.ChangeEvent<HTMLSelectElement>) =>
                             setFormVariables({
@@ -214,6 +214,7 @@ export const AbuseReportForm: React.FC<{
                     </label>
                     <textarea
                         name="reason"
+                        id="reason"
                         onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) =>
                             setFormVariables({
                                 ...formVariables,
@@ -236,6 +237,7 @@ export const AbuseReportForm: React.FC<{
                     <input
                         type="email"
                         name="email"
+                        id="email"
                         onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
                             setFormVariables({
                                 ...formVariables,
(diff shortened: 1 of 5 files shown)


gtrufitt:
Ideally we'd pull in the styling from https://theguardian.design/2a1e5182b/p/097455-text-input-field/b/050445/t/663407 but I think this is ok for now

gtrufitt:
[Comment reporting not working](https://trello.com/c/KMFCdRWe/1553-comment-reporting-not-working)

gtrufitt:
> Do you think it's worth trying to add a couple of tests for this? Not sure if this would be easy or not but we're a bit light on unit tests for this lib so it'd be great to start building that up

Tests added @oliverlloyd - happy for you to look (and merge if you really want!:D)

src/components/AbuseReportForm/AbuseReportForm.tsx
```diff
@@ -135,47 +135,48 @@ export const AbuseReportForm: React.FC<{
...
-                    </span>
-                )}
-
                 <div className={inputWrapper}>
                     <label className={labelStylesClass} htmlFor="category">
                         Category
                     </label>
                     <select
-                        name="categoryId"
+                        name="category"
```
oliverlloyd, line 179: This change is breaking some of the accessiblity features. I agree though that `categoryId` is a bad name but if we change it we should carry that through the code

gtrufitt: > This change is breaking some of the accessiblity features. I agree though that `categoryId` is a bad name but if we change it we should carry that through the code

I actually changed it because it didn’t match the label for attribute previously... I’ll double check.

gtrufitt: In fact, I’m fairly sure this is correct and I noticed because clicking the label didn’t select the field before this change!

oliverlloyd: I made [a change](https://github.com/guardian/discussion-rendering/pull/249/commits/edb4ac9e492349a8d415dbad769f0ca4c4ea82fe) that I think fixes this. The code to override the tabbing was expecting the name to be `categoryId` but I changed it to `category`.

gtrufitt: Ah perfect thanks!

PR #250, opened 2020-05-04
Show filters when unexpanded

oliverlloyd:
## What does this change?
Show filters, even if the comments are not expanded

Screenshot 2020-05-04 at 11 46 58

### Test updates
This PR also refactors the tests to include scope for this change and also to better test top picks in general

## Why?
When data is listed, even limited data, we should also give the reader the controls that apply to that list


oliverlloyd:
[Discussion missing filters](https://trello.com/c/TUVkb6O3/1568-discussion-missing-filters)

PR #255, opened 2020-05-12
Fix show more replies

liywjl:
## What does this change?
change `displayResponse` to `displayResponses` in the query param

## Why?
because responses do not get fetched if so

https://trello.com/c/MtfHvXLz/1654-show-more-replies-broken

## To Test
`displayResponses`
https://discussion.theguardian.com/discussion-api/comment/104345678?displayThreaded=true&displayResponses=true

`displayResponse`
https://discussion.theguardian.com/discussion-api/comment/104345678?displayThreaded=true&displayResponse=true

### DIFF
--- a/src/components/CommentContainer/CommentContainer.tsx
+++ b/src/components/CommentContainer/CommentContainer.tsx
@@ -130,7 +130,7 @@ export const CommentContainer = ({
         getMoreResponses(commentId)
             .then(json => {
                 setExpanded(true);
-                setResponses([...responses, ...(json.comment.responses || [])]);
+                setResponses(json.comment.responses || []);
             })
             .finally(() => {
                 setLoading(false);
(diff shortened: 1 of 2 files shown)


src/components/CommentContainer/CommentContainer.tsx
```diff
@@ -130,7 +130,7 @@ export const CommentContainer = ({
         getMoreResponses(commentId)
             .then(json => {
                 setExpanded(true);
-                setResponses(json.comment.responses || []);
+                setResponses([...responses, ...(json.comment.responses || [])]);
```
oliverlloyd, line 133: I think the problem here is that `json.comment.responses` is not (no longer?) being returned. I'm not sure that adding `...responses` will provide the replies.

PR #257, opened 2020-05-14
onHeightChange

oliverlloyd:
## What does this change?
Here we add a callback to be fired each time the height of the comments container changes

## Why?
In DCR, we want to know when the contents might be causing the height of our container to change so that we can adjust the css settings for the `StickyAd`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,6 +38,7 @@ type Props = {
     expanded: boolean;
     onPermalinkClick: (commentId: number) => void;
     apiKey: string;
+    onHeightChange: () => void;
 };
 
 const footerStyles = css`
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[onHeightChange](https://trello.com/c/zhJqiPaO/1662-onheightchange)

src/App.tsx
```diff
@@ -38,6 +38,7 @@ type Props = {
     expanded: boolean;
     onPermalinkClick: (commentId: number) => void;
     apiKey: string;
+    onHeightChange: () => void;
```
gtrufitt, line 41: Should it be mandatory?

oliverlloyd: Tricky question. 

No. If we want discussion to work in other contexts
Yes. If we assume discussion is entwined with DCR

I guess my head is assuming the entwine but we should strive for the separation. I'll update

PR #258, opened 2020-05-14
Expand on post

oliverlloyd:
## What does this change?
If the view is not expanded then when a reader posts we now expand.

At the same time,  we also scroll the page to show the post just made.

## Why?
Because if a reader is signed in and the view not yet expanded they are still able to post and at this point they are engaged and we should take them to the main content.


oliverlloyd:
[Expand comments on post](https://trello.com/c/2XO1RB90/1660-expand-comments-on-post)

PR #259, opened 2020-05-14
onHeightChange is now optional

oliverlloyd:
## What does this change?
Makes the `onHeightChange` callback optional

## Why?
Because we don't want to be too proscriptive about how discussion should be used


oliverlloyd:
[onHeightChange optional](https://trello.com/c/PZf7Q52u/1663-onheightchange-optional)

PR #260, opened 2020-05-14
Scroll to form on reply

oliverlloyd:
## What does this change?
Now when you click reply, the form that gets shown is scrolled into view

## Why?
Because sometimes the list of replies can be quite long and the form can get lost


oliverlloyd:
[Scroll to reply form](https://trello.com/c/HqgEj9l2/1658-scroll-to-reply-form)

PR #261, opened 2020-05-14
Preview

oliverlloyd:
## What does this change?

Screenshot 2020-05-14 at 17 09 02
Adds a Preview component with the correct styles

## Why?
We had overlooked this because we weren't testing for it


oliverlloyd:
[Fix Preview](https://trello.com/c/VqXbgp3V/1619-fix-preview)

PR #270, opened 2020-05-22
Comments are unpicked by default

oliverlloyd:
## What does this change?
Ensures that new comments do not appear as picked when they are first posted (When simulating a comment to display optimistically, the `isHighlighted` property was being set to true).

## Why?
Because they're not


oliverlloyd:
[Highlighting the wrong comments as 'picked'](https://trello.com/c/sY328IMa/1650-highlighting-the-wrong-comments-as-picked)

PR #272, opened 2020-05-22
Ensure comment styles are correct

oliverlloyd:
## What does this change?
Reviews and enforces specific styles for comments in both the comment and preview views

## Why?
A reader raised a problem where their comment text was not correctly being shown as italic. Upon investigation, I also noticed that some other aspects of formnatting were not always being set properly so here I fixed:

- strikethrough tag set wrongly (`strike`, not `del`)
- code formtting missing
- bold formtting missing
- italic formtting missing

### Preview before
Screenshot 2020-05-22 at 14 56 15
### Preview after
Screenshot 2020-05-22 at 12 28 12

### Comment before
Screenshot 2020-05-22 at 14 48 51

### Comment after
Screenshot 2020-05-22 at 15 37 26

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -88,6 +88,20 @@ const commentCss = css`
     padding-left: ${space[2]}px;
     color: ${neutral[46]};
   }
+
+  i {
+      font-style: italic;
+  }
+
+  b {
+      font-weight: bold;
+  }
+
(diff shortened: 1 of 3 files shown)


oliverlloyd:
[i button doesn't create italics](https://trello.com/c/iz3Ru5di/1691-i-button-doesnt-create-italics)

PR #276, opened 2020-05-27
Is a discussion is closed, default orderBy to oldest

oliverlloyd:
## What does this change?
Changes the default `orderBy` value from always `newest` to use a function based on if the discussion is cloded or not

## Why?
Because  this is how frontend does this and readers will be used to this


oliverlloyd:
[Default to orderBy oldest](https://trello.com/c/VIcL6ZYY/1711-default-to-orderby-oldest)

PR #277, opened 2020-05-27
Default to 100 comments per page

oliverlloyd:
## What does this change?
Changes the default count of comments to show from 25 to 100

## Why?
We don't want to hide content and force readers to needlessly scroll


oliverlloyd:
[Default to pageSize 100](https://trello.com/c/vXH8nDWN/1710-default-to-pagesize-100)

PR #281, opened 2020-06-09
Add nofollow to all links

gtrufitt:
## What does this change?

Adds `rel="nofollow"` to all comment links.

## Why?

In Frontend we block these links with robots.txt. We haven't replicated that in DCR.

https://trello.com/c/jgohtyEQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@guardian/discussion-rendering",
     "description": "This codebase started as a hack day project by @gtrufitt and @nicl. The purpose is parity of the existing discussion application on Frontend using the discussion API (search for Private Repo).",
-    "version": "2.2.22",
+    "version": "2.2.23",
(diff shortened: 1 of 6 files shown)


oliverlloyd:
There are some links that show in info/warning messages that get surfaced when we get certain api responses.

Eg: 

```
...
            } else if (response.message === 'USER_BANNED') {
                setError(
                    'Commenting has been disabled for this account (<a href="/community-faqs#321a">why?</a>).',
                );
            } else if (response.message === 'IP_THROTTLED') {
                setError(
                    'Commenting has been temporarily blocked for this IP address (<a href="/community-faqs">why?</a>).',
                );
...
```

Do we need to include those?

gtrufitt:
> Do we need to include those?

Not too fussed, we could do but I don't think they matter do much being a) internal b) single pages.

PR #291, opened 2020-06-23
Add specific styling to picked comment links

jfsoul:
## What does this change?
Adds specific styling (no underline until hover, and different blue font) to links within picked comment bodies.

## Why?
To better match the styling of picked comment links on frontend (although we have been consistent with normal discussion rendering comment links rather than blindly copy frontend).

Before:
image

After:
image

https://trello.com/c/eHLPC8fm/1693-pick-link-colour-is-incorrect


src/components/TopPick/TopPick.tsx
```diff
@@ -67,6 +67,17 @@ const linkStyles = css`
     }
 `;
 
+// to override a tag styles from dangerouslySetInnerHTML
+const inCommentLinkStyling = css`
+    a {
+        color: ${palette.brand[500]};
```
oliverlloyd, line 73: We moved away from using the `palette` object to try and reduce bundles size. You can get `brand` with 

```
import { brand } from '@guardian/src-foundations/palette';
```

Obvs, in this case, `palette` is already imported but that should, ideally, be refactored away at some point so it's good to stick to the convention to make that refactor easier

src/components/Comment/Comment.stories.tsx
```diff
@@ -7,8 +7,7 @@ export default { title: 'Comment' };
 
 const commentData: CommentType = {
     id: 25487686,
-    body:
-        "<p>Beau Jos pizza in Idaho Springs is a great place for mountain pizza pies. Order one with extra thick crust and drizzle it with honey. Y'all can try the Challenge if you fancy, and sketch on your napkins so your art can join their walls. This was 15 years ago, but I hope it's still there! As for music, anything from Boulder's own Big Head Todd &amp; the Monsters - 'Broken Hearted Savior' is a good start, with 'Bittersweet' a good road track. I'm jealous!!!</p>",
+    body: `<p>Beau Jos pizza in Idaho Springs is a great place for <a href="https://www.theguardian.com">mountain pizza pies</a>. Order one with extra thick crust and drizzle it with honey. Y'all can try the Challenge if you fancy, and sketch on your napkins so your art can join their walls. This was 15 years ago, but I hope it's still there! As for music, anything from Boulder's own Big Head Todd &amp; the Monsters - 'Broken Hearted Savior' is a good start, with 'Bittersweet' a good road track. I'm jealous!!!</p>`,
```
oliverlloyd, line 10: 👍

src/components/TopPick/TopPick.tsx
jfsoul: Thanks, have updated!

jfsoul:
P.s. @oliverlloyd I was planning to release this to npm, but as it's non-urgent I am tempted to let it go out with a few other things - or just whenever the next critical change is. Objections?

oliverlloyd:
> P.s. @oliverlloyd I was planning to release this to npm, but as it's non-urgent I am tempted to let it go out with a few other things - or just whenever the next critical change is. Objections?

No objections either way. I think the discussion on how and how often to release is fairly open at the moment, it's a bit of a chore how we have things setup right now but perhaps with more tooling (maybe a monorepo?) it could become so easy as to be routine.

PR #316, opened 2020-07-31
Ensure that long usernames don't push out replies

gtrufitt:
## What does this change?

Usernames will be truncated if they are too long and don't have enough space, instead of before where they pushed the comments container too wide.

They truncate in the most space effective way possible, always showing as much of each username as is practical.

### Before
image

### After
Screen Shot 2020-07-31 at 11 35 38
Screen Shot 2020-07-31 at 11 35 30
Screen Shot 2020-07-31 at 11 34 46

### DIFF
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -233,7 +233,7 @@ const cssReplyToWrapper = css`
 // - Both labels should truncate with ellipsis if they fill their space
 // Test page: https://codepen.io/gtrufitt/pen/LYGKQyY
 
-const cssReplyDisplayName = css`
+const cssReplyAlphaDisplayName = css`
     ${until.mobileLandscape} {
         ${cssTextOverflowElip}
         width: 100%;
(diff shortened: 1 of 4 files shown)


src/components/Comment/Comment.tsx
```diff
@@ -211,6 +212,43 @@ const negativeMargin = css`
...
+
+const cssReplyDisplayName = css`
+    ${until.mobileLandscape} {
+        ${cssTextOverflowElip}
+        width: 100%;
+        max-width: fit-content;
+    }
+`;
+
+const cssReplyToDisplayName = css`
```
jfsoul, line 244: I guess these are accurate names, but a bit too similar at first glance!

src/index.tsx
```diff
@@ -38,6 +38,7 @@ const IndexPageWrapper = () => {
                 </a>
                 .
             </p>
+            <style>{`body { padding: 0 10px;}`}</style>
```
jfsoul, line 41: This necessary to avoid the overflow or just an additional unrelated change?

gtrufitt: Yeah, unrelated, the index.tsx is used *just* for local testing (it doesn't get bundled for the app)

PR #344, opened 2020-10-08
on recommend callback

webb04:
## What does this change?
The apps don't have access to the user cookie required to validate a number of actions like posting, replying, recommending or reporting.

For these events, discussion-rendering could accept optional callbacks as props. This would allow native apps to make the requests with their identity tokens and report back the success/failure to the webview.

The Bridget communication is [promise based](https://github.com/guardian/apps-rendering/blob/c66f6db84c0af623a07d2b7a96b231f92da6c135/src/client/nativeCommunication.ts#L82) so I'm hoping they could just be passed through as props.

I've added default props to keep the current behaviour, so dotcom-rendering shouldn't need to add any changes.

This is a draft PR that can be extended if needed.

### DIFF
--- a/src/components/RecommendationCount/RecommendationCount.tsx
+++ b/src/components/RecommendationCount/RecommendationCount.tsx
@@ -15,6 +15,7 @@ type Props = {
     alreadyRecommended: boolean;
     isSignedIn: boolean;
     userMadeComment: boolean;
+    onRecommend: (commentId: number) => Promise<Boolean>;
 };
 
 const countStyles = css`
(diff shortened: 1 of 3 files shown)


src/components/RecommendationCount/RecommendationCount.tsx
```diff
@@ -15,6 +15,7 @@ type Props = {
     alreadyRecommended: boolean;
     isSignedIn: boolean;
     userMadeComment: boolean;
+    onRecommend: (commentId: number) => Promise<Boolean>;
```
JamieB-gu, line 18: Would this have to be an optional prop in order for dotcom to ignore it and have a fallback to the default?

```suggestion
    onRecommend?: (commentId: number) => Promise<Boolean>;
```

JamieB-gu (commented):
This seems like a nice approach to me 👍

gtrufitt (approved):
Nice solution!

liywjl (approved):
LGTM 👍

PR #362, opened 2020-12-21
Dynamic pagination text

oliverlloyd:
## What does this change?
Updates Pagination to change the text it uses based on the reader's filter preferences.

Say 'Displaying threads' when we're counting by threads and say 'Displaying comments' when we're counting by comments

## Why?
It's actually really confusing if we always show the text 'Displaying comments 1 to 203' because actually we're not always doing that. If the reader's filter prefs for threading are set to something other than `unthreaded` then it means we 'count' things based on threads, not the number of actual comments.

This means the total count will change. A discussion with 2 comments that both have 10 replies will have a total of 22 comments but only 2 'threads'. 

Screenshot 2020-12-21 at 13 05 55

### DIFF
--- a/src/components/Pagination/Pagination.stories.tsx
+++ b/src/components/Pagination/Pagination.stories.tsx
@@ -136,3 +136,45 @@ export const LotsOfPages = () => {
     );
 };
 LotsOfPages.story = { name: 'with many pages' };
+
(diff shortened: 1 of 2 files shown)


liywjl (approved):
👍

PR #373, opened 2021-01-05
Update source v2.7.1

liywjl:
## What does this change?

- Update source packages to v2.7.1
- Make fixes to Source overrides

## Why?
- DCR is using source v2.7.1
- Source will be bumping to v3 soon, so its best to get ready for an update

https://trello.com/c/3MZnq43Z/2281-comments-not-properly-fitting-on-the-screen-on-iphones
https://trello.com/c/byuuGWVD/2289-update-source-in-discussion-rendering

## Next steps
Implement the removed override style changes in Source


src/components/Comment/Comment.tsx
```diff
@@ -408,6 +413,7 @@ export const Comment = ({
 													regularFont,
 													svgOverrides,
 													cssReplyBetaDisplayName,
+													linkOverrides,
```
oliverlloyd, line 416: Is this version of Source introducing a new problem that we need to now fix?

src/components/Comment/Comment.tsx
```diff
@@ -168,14 +164,23 @@ const regularFont = css`
...
+		display: flex;
+		flex-direction: row;
+		align-items: inherit;
+	}
+`
+
+const buttonOverrides = css`
+	.src-button-space {
+		width: 0px !important;
+	}
```
oliverlloyd: This coupling to classnames from source with !important on top probably isn't something we want to do. Maybe we can just accept the way Souorce does things?

src/components/Comment/Comment.tsx
```diff
@@ -529,20 +535,19 @@ export const Comment = ({
...
 												{user ? (
-													<div className={svgOverrides} id="fgsghdiusfhfdsiu">
+													<div className={cx(svgOverrides, buttonOverrides)}>
 														<ButtonLink
 															pillar={pillar}
 															onClick={() => setCommentBeingRepliedTo(comment)}
 															icon={<SvgIndent />}
 															iconSide="left"
-															linkName="reply to comment"
+															linkName="reply-to-comment"
```
oliverlloyd: This change seems unrelated. What's the reasoning for this?

oliverlloyd (commented):
👍 for the removal of the existing overrides! But it seems like we are also introducing a lot more 😞

liywjl:
superseded by https://github.com/guardian/discussion-rendering/pull/374

src/components/Comment/Comment.tsx
liywjl: lots of the override styles seem not to work anymore with updated source. I think it might be worth taking a step back and trying to find a more stable work around for things we do not feel are right

liywjl:
https://github.com/guardian/discussion-rendering/pull/374 is a prerelease and many of the issues raised here seem to persiste (hopefully fixing them here fix them there as well)

src/components/Comment/Comment.tsx
SiAdcock: Let's chat! 🗣️

src/components/Comment/Comment.tsx
buck06191: I think @liywjl preferred the linkName to be hyphenated over having spaces. I think it makes sense as that should keep it in line with other link names, but potentially it could be pulled out as a separate PR to keep the diff clean

.vscode/settings.json
```diff
@@ -14,5 +14,6 @@
   "eslint.enable": true,
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib"
```
jfsoul, line 18: What's this for?

src/components/Comment/Comment.tsx
```diff
@@ -531,69 +522,65 @@ export const Comment = ({
...
+										<>
+											{/* If user is not logged in we link to the login page */}
+											{user ? (
+												<div className={svgReplyArrow}>
+													<ButtonLink
+														pillar={pillar}
+														onClick={() => setCommentBeingRepliedTo(comment)}
+														icon={<SvgIndent />}
+														iconSide="left"
+														linkName="reply-to-comment"
```
jfsoul, line 537: is this change important? unlikely but it might break a report somewhere

jfsoul (approved):

.vscode/settings.json
buck06191: This probably needs reverting. My VSCode kicked up a fuss and wouldn't use the correct typescript version without this. It basically tells VSCode to use the project typescript version instead of the system version. I can remove it, or if it seems useful it can stay?

src/components/Comment/Comment.tsx
buck06191: If you mean the link name, I think this can be reverted and was mainly something @liywjl implemented to make it consistent with the rest of the link names.

.vscode/settings.json
oliverlloyd: I made some changes to the ts version recently which is probably related to this. Shall we move this change into a separate PR though so we can figure out if we need it set at this level or not? I'm not really sure at the moment to be honest and I don't think we should hold up this Source change for it

PR #383, opened 2021-01-07
Fix bug with long usernames

buck06191:
## What does this change?
Long usernames were causing the width to stretch in mobile view on DCR due to the width not being set on a container. This moves from using a `Column` for comments, to a div with the same flex styling but with a `max-width` of `100%`.

I've also added a story to explicitly check for threaded replies with long usernames, though in this case that wasn't causing an issue as it was a combination of factors that only occurred when implementing in DCR.

Also requires https://github.com/guardian/dotcom-rendering/pull/2342 to be merged.

## Why?
Poor UX when expanding comments. Long usernames caused horizontal scroll on mobile views.

https://trello.com/c/3MZnq43Z/2281-comments-not-properly-fitting-on-the-screen-on-iphones

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -56,6 +56,12 @@ const footerStyles = css`
 	justify-content: flex-end;
 `;
 
+const commentColumnWrapperStyles = css`
+	display: flex;
+	flex-direction: column;
+	max-width: 100%;
+`;
+
 const commentContainerStyles = css`
 	display: flex;
 	flex-direction: column;
@@ -460,7 +466,7 @@ export const App = ({
 	}
 
 	return (
-		<Column>
+		<div className={commentColumnWrapperStyles}>
 			<div data-component="discussion">
 				{user && !isClosedForComments && (
 					<CommentForm
@@ -553,6 +559,6 @@ export const App = ({
 					/>
 				)}
 			</div>
-		</Column>
+		</div>
 	);
 };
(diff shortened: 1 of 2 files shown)


buck06191:
[Comments not properly fitting on the screen on iPhones](https://trello.com/c/3MZnq43Z/2281-comments-not-properly-fitting-on-the-screen-on-iphones)

src/App.tsx
```diff
@@ -460,7 +466,7 @@ export const App = ({
 	}
 
 	return (
-		<Column>
+		<div className={commentColumnWrapperStyles}>
```
oliverlloyd, line 469: Now that we're not using Column, I wonder if we can merge this div with its child? One less div is always a win

oliverlloyd (approved):
Impressive detective work getting to the bottom of this, and 💯 for the story!

I left one comment wondering if there was a chance to consolidate 2 divs into one and I can see Github is asking for the unused Column import to be removed but other than that 👍

oliverlloyd:
Just a thought. It might be nice to set a [smaller viewport width on the chromatic](https://www.chromatic.com/docs/viewports) snapshot so we diff against how this component renders on mobile with a long username

liywjl (approved):
Great work!!! This really was a difficult one to crack.

You can remove the unused import in `App.tsx`

Bar that all looks great 👍

PR #384, opened 2021-01-07
Upgrade babel core and preset typescript

SiAdcock:
## What does this change?

- upgrade @babel/preset-typescript
- upgrade @babel/core

## Why?

`@guardian/types` [uses the export type syntax](https://github.com/guardian/types/blob/cd4f43cac0e20c63b69c08c2a0e34602ea1455c3/src/format.ts#L51) that was [introduced in TypeScript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export). Support for this was added to Babel's TypeScript preset in [v7.9.0](https://github.com/babel/babel/releases/tag/v7.9.0).

Storybook fails to load without these upgrades

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,11 +40,11 @@
     "typescript": "^4.1.3"
   },
   "devDependencies": {
-    "@babel/core": "^7.5.5",
+    "@babel/core": "^7.12.10",
     "@babel/plugin-transform-runtime": "^7.9.0",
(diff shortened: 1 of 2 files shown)


oliverlloyd (approved):
Thanks 👏

PR #385, opened 2021-01-07
Disable focus manager on click in storybook

SiAdcock:
## What does this change?

Disables Source's [Focus Manager](https://theguardian.design/2a1e5182b/p/6691bb-accessibility/t/32e9fb) unless a tab keypress is detected.

## Why?

By default, Source components have a blue focus "halo" which is visible on keyboard focus, and also on click interaction. It is undesirable to show this halo on click. The Focus Manager must be disabled manually.

**Before**

**After**

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,6 +1,4 @@
-import React from "react"
-import { addDecorator } from "@storybook/react"
-import { useEffect } from "@storybook/addons"
+import React, { useEffect } from "react"
 import { FocusStyleManager } from "@guardian/src-foundations/utils"
 import MockDate from 'mockdate';
 import { mockFetchCalls } from '../src/lib/mockFetchCalls';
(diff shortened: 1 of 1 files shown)


.storybook/preview.js
```diff
@@ -1,3 +1,7 @@
+import React from "react"
+import { addDecorator } from "@storybook/react"
+import { useEffect } from "@storybook/addons"
```
oliverlloyd, line 3: Well, this is a thing I didn't know about

oliverlloyd (approved):
👍

.storybook/preview.js
SiAdcock: Good point, I can't even find the docs I copy-pasted this from. 

Why not use React's `useEffect`? It seems to work fine! I have updated this accordingly

liywjl (approved):

PR #389, opened 2021-01-08
Types @ v1.1

oliverlloyd:
## What?
Bumps @guardian/type to v `1.1.0`

## What else?
Adds `ts-jest` lib, converting  js files to ts and removing usage of optional chaining and nullesh colascing

## Why?
We want access to the `PhotoEssay` design type in DCR and so we need to update this package in each dependency


src/lib/dateFormatter.ts
```diff
@@ -16,13 +16,13 @@ const monthConverter = {
 	11: 'Dec',
 };
 
-const isLast24Hrs = (date) => {
+const isLast24Hrs = (date: any) => {
```
buck06191, line 19: Is there a reason to use `any` instead of `Date` here? Given that we then do a `>=` with it against a `Date` object, does it not need to also be a `Date` object?

liywjl (approved):
LGTM 👍  some CI issues to resolve however

oliverlloyd:
[@guardian/types v1.1 for discussion-rendering](https://trello.com/c/Ec93nS6W/2311-guardian-types-v11-for-discussion-rendering)

src/lib/dateFormatter.ts
oliverlloyd: Good point. I started changing this but it surfaced an issue where this function might not be working so I'm going to raise a ticket for it

oliverlloyd: https://trello.com/c/6aZX3gEr/2322-islast24hours-function-in-discussion

oliverlloyd: I added a fix for this

PR #390, opened 2021-01-13
Add np lib to publish changes

oliverlloyd:
## What does this change?
Replaces our custom script used for publishing with the [np library](https://github.com/sindresorhus/np).

```bash
yarn release // Runs validation, builds and then opens an interactive UI to pubish changes
```


liywjl (commented):
🎊

PR #397, opened 2021-01-21
Use Theme, not Pillar

oliverlloyd:
## What does this change?
1. Replaces `Pillar` with `Theme`
2. Bumps @guardian/types to v3.0.0

## Why?
To enable support of `SpecailReport` and `Labs` and we want to bring through `PrintShop` design type from /types


oliverlloyd:
[Use Theme in discussion-rendering](https://trello.com/c/5r0eYD5r/2363-use-theme-in-discussion-rendering)

liywjl (approved):
Do we want to change the prop names from pillar to theme?

oliverlloyd:
> Do we want to change the prop names from pillar to theme?

I did think about this but I think for that level of change we need to give it some thought. CAPI is still using `pillar` and it could be confusing to transform from this.

Better might be to move entirely over to just using `format: Format` and remove the `pillar` prop entirely! (And use `format.theme` instead)

gtrufitt:
> Better might be to move entirely over to just using format: Format and remove the pillar prop entirely

Looking at the DCR PR, definitely feels like we need to move to Format fairly soon as we have Pillar and Theme which is defo confusing!

PR #401, opened 2021-01-28
Errors appearing in Discussion Rendering

nitro-marky:
## What does this change?

Changes `response.message` to `response.errorCode` to fix Discussion errors. Stories have also been added to confirm this works.

This is what happens when a user is not registered:

Screenshot 2021-01-28 at 11 59 38


oliverlloyd:
This is odd. So did the response format from the discussion api _change_? I'd be keen to keen to see that PR here if so. Or did it _never_ send this field (yikes).

_Or_ does it sometimes send one and sometimes the other...?

gtrufitt:
> This is odd. So did the response format from the discussion api _change_? I'd be keen to see that PR here if so. Or did it _never_ send this field (yikes).
> 
> _Or_ does it sometimes send one and sometimes the other...?

It has been `errorCode` for at least 7 years 🤷‍♂️  https://github.com/guardian/discussion-api/blame/8b46892263738b7d9be5cd4f98dd8c8d9eb92f41/discussion-api/src/main/scala/com.gu.discussion.api/model/Response.scala#L26

nitro-marky:
Is this good to merge?

src/components/CommentForm/CommentForm.stories.tsx
```diff
@@ -66,6 +66,16 @@ export const Default = () => (
...
 
+export const Error = () => (
+	<CommentForm
+		shortUrl={'/p/g8g7v'}
+		pillar={Pillar.News}
+		user={aUser}
+		onAddComment={(comment) => {}}
+	/>
+);
+Error.story = { name: 'Error' };
```
buck06191, line 77: Can we change the name to provide more info, similar to `Active` below it? 
```suggestion
Error.story = { name: 'form with errors' };
```

buck06191 (approved):
It LGTM, but I don't know if the above comments about where this error has recently come from might be worth doing a bit of investigation. Could it be that this is an error that has always been present but not noticed?

Now that we have stories for it, we should catch this more often so great work on that.

oliverlloyd (approved):
> Is this good to merge?

So I checked the [discussion errors](https://github.com/guardian/discussion-api/blob/f05bd83ef992c11e1c4eb170d9426123c1a13c1d/discussion-api/src/main/scala/com.gu.discussion.api/api/ApiRequestException.scala) (`errorCode`), and I checked what [Frontend uses](https://github.com/guardian/frontend/blob/d7d261ac608a5b1c27a5977a5ef70b978cce91c4/static/src/javascripts/projects/common/modules/discussion/comment-box.spec.js#L50) (also `errorCode`) so yeah, we're good to merge 🤯

src/components/CommentForm/CommentForm.stories.tsx
```diff
@@ -66,6 +66,16 @@ export const Default = () => (
 );
 Default.story = { name: 'default' };
 
+export const Error = () => (
```
gtrufitt, line 69:
```suggestion
// This story has a mocked post endpoint that returns an error, see 97d6eab4a98917f63bc96a7ac64f7ca7
export const Error = () => (
```

src/lib/mockFetchCalls.ts
```diff
@@ -143,6 +143,19 @@ export const mockFetchCalls = () => {
 			status: 200,
 		})
 
+		//https://discussion.theguardian.com/discussion-api/discussion/p/g8g7v/comment?api-key=dotcom-rendering
```
gtrufitt, line 146:
```suggestion
		//https://discussion.theguardian.com/discussion-api/discussion/p/g8g7v/comment?api-key=dotcom-rendering
		// This returns a mocked error for a story 97d6eab4a98917f63bc96a7ac64f7ca7
```

gtrufitt (approved):

oliverlloyd:
Probably worth pointing out that this highlights a gap in our test suite. We ideally want to be able to run an end to end integration test where we actually talk to the discussion api and verify the response we get back

gtrufitt:
@nitro-marky There shouldn't be a package-lock I don't think?

nitro-marky:
According to the file `32,538 additions, 0 deletions not shown because the diff is too large. Please use a local Git client to view these changes.`

gtrufitt:
> According to the file `32,538 additions, 0 deletions not shown because the diff is too large. Please use a local Git client to view these changes.`
> 
> Tried recreating the situation without any luck. I can delete this file in the PR and push again and see what happens 🤷‍♂️

package-lock is for npm, I imagine you accidentally ran `npm` instead of `yarn`? So yeah, deleting is the way to go!

PR #402, opened 2021-01-28
Update settings to add titlebar colour and fix extensions

buck06191:
## What does this change?
Changes title bar colour to aid in between project distinctions. colour palette taken from https://davidmathlogic.com/colorblind/#%23648FFF-%23785EF0-%23DC267F-%23FE6100-%23FFB000
image

## Why?
Because I get confused easily


oliverlloyd (approved):
discussion-rendering always did feel pink to me

PR #409, opened 2021-01-29
 Update to @emotion/core

liywjl:
## What does this change?
Update the use of emotion to be more in line with how you write emotion 10.

## Why?
We have been using Emotion 10 but we write emotion code as if we are using Emotion 9 (this was allowed to help make the transition simple). But now as we are aiming to go to Emotion 11, we should first migrate to using Emotion 10 syntax.


liywjl:
superseded by https://github.com/guardian/discussion-rendering/pull/411

PR #411, opened 2021-02-01
Migrate to @emotion/core

liywjl:
## What does this change?
Update the use of `@emotion/core` to be more in line with how you write emotion 10.

## Why?
We have been using Emotion 10 but we write emotion code as if we are using Emotion 9 (this was allowed to help make the transition simple). But now as we are aiming to go to Emotion 11, we should first migrate to using Emotion 10 syntax.

## JSX?
Emotion uses `jsx` exposed from `@emotion/core` (in Emotion 10), rather than React, as a way of converting `css` props on components to `classNames`. In order to get this working you need to add a babel preset `@emotion/babel-preset-css-prop` and make sure that react files start with `/** @jsx jsx */` (a way to indicate to babel that you want to use JSX rather than React.

## Fragments?
Babel does not understand `<></>`, therefore we use React to explicitly define `React.Fragment`

## JSX Runtime?
We can only support JSX runtime classic untile we migrate to Emotion 11, so each file needs to have the pragma:`/** @jsxRuntime classic */` at the top

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -25,7 +25,7 @@ const guardianViewports = {
   phablet: {
       name: 'phablet',
       styles: {
-          width: '660px',
+          width: '660px', 
           height: '800px',
       },
   },
--- a/babel.config.js
+++ b/babel.config.js
@@ -15,4 +15,9 @@ module.exports = {
         "@babel/preset-react",
         '@emotion/babel-preset-css-prop'
     ],
+    env: {
+        test: {	
+            plugins: ['@babel/plugin-transform-runtime'],	
+        },	
+    },
 };
--- a/babel.config.js
+++ b/babel.config.js
@@ -2,12 +2,7 @@ module.exports = {
     plugins: [
         'const-enum',
         '@babel/transform-typescript',
-        [
-            'babel-plugin-emotion',
-            {
-                'cssPropOptimization': true
-            }
-        ]
+        'babel-plugin-emotion',
     ],
     presets: [
         '@babel/preset-env',
--- a/babel.config.js
+++ b/babel.config.js
@@ -16,8 +16,8 @@ module.exports = {
         '@emotion/babel-preset-css-prop'
     ],
     env: {
-        test: {	
-            plugins: ['@babel/plugin-transform-runtime'],	
-        },	
+        test: {
+            plugins: ['@babel/plugin-transform-runtime'],
+        },
     },
 };
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,14 +1,18 @@
 module.exports = {
-    plugins: ["const-enum", "@babel/transform-typescript"],
+    plugins: [
+        "const-enum",
+        "@babel/transform-typescript",
+        [
+            "babel-plugin-emotion",
+            {
+                "cssPropOptimization": true
+            }
+        ]
+    ],
     presets: [
         '@babel/preset-env',
         '@babel/preset-typescript',
-        '@babel/preset-react',
-        '@emotion/babel-preset-css-prop',
+        "@babel/preset-react",
+        '@emotion/babel-preset-css-prop'
     ],
-    env: {
-        test: {
-            plugins: ['@babel/plugin-transform-runtime'],
-        },
-    },
 };
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,18 +1,18 @@
 module.exports = {
     plugins: [
-        "const-enum",
-        "@babel/transform-typescript",
+        'const-enum',
+        '@babel/transform-typescript',
         [
-            "babel-plugin-emotion",
+            'babel-plugin-emotion',
             {
-                "cssPropOptimization": true
+                'cssPropOptimization': true
             }
         ]
     ],
     presets: [
         '@babel/preset-env',
         '@babel/preset-typescript',
-        "@babel/preset-react",
+        '@babel/preset-react',
         '@emotion/babel-preset-css-prop'
     ],
     env: {
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "timeago.js": "^4.0.2"
   },
   "peerDependencies": {
-    "@emotion/core": "^10.0.28",
+    "@emotion/core": "10.0.35",
     "@guardian/src-button": "2.7.1",
     "@guardian/src-foundations": "2.7.1",
     "@guardian/src-icons": "2.7.1",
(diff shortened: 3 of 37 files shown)


shtukas (approved):

babel.config.js
```diff
@@ -1,10 +1,19 @@
...
+                "cssPropOptimization": true
+            }
+        ]
+    ],
     presets: [
         '@babel/preset-env',
         '@babel/preset-typescript',
-        '@babel/preset-react',
-        '@emotion/babel-preset-css-prop',
+        "@babel/preset-react",
```
SiAdcock, line 15: lol them quotes

buck06191:
Does this not work with regards to Babel and Fragments: https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#fragments

buck06191:
Does this not work with regards to Babel and Fragments: https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#fragments ?

babel.config.js
```diff
@@ -1,10 +1,19 @@
 module.exports = {
-    plugins: ["const-enum", "@babel/transform-typescript"],
+    plugins: [
+        "const-enum",
+        "@babel/transform-typescript",
+        [
+            "babel-plugin-emotion",
+            {
+                "cssPropOptimization": true
```
SiAdcock, line 8: [This is a default](https://emotion.sh/docs/@emotion/babel-plugin#csspropoptimization), I don't think it needs to be explicit

SiAdcock:
Is it safe to remove the [`emotion`](https://github.com/guardian/discussion-rendering/pull/411/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L70) dependency?

src/App.stories.tsx
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
+/** @jsxRuntime classic */
+/** @jsx jsx */
```
buck06191, line 2: We should only need the pragma if we're [not using the `babel-preset-css-prop` plugin](https://emotion.sh/docs/css-prop#jsx-pragma)

Do you know why we need both here?

src/App.stories.tsx
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
+/** @jsxRuntime classic */
```
SiAdcock, line 1: I think this is only required from [Babel 7.9.0](https://babeljs.io/docs/en/babel-preset-react/). Do you plan on updating that?

buck06191, line 1: Can we use [`@babel/plugin-transform-react-jsx`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#usage) to avoid this by setting the run time at a higher level?

src/components/Avatar/Avatar.stories.tsx
```diff
@@ -5,7 +5,7 @@ export default { component: Avatar, title: 'Avatar' };
 
 export const Sizes = () => {
 	return (
-		<>
+		<React.Fragment>
```
buck06191, line 8: As above, can we use[`@babel/plugin-transform-react-jsx
`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#usage) to avoid this?

babel.config.js
```diff
@@ -1,10 +1,19 @@
 module.exports = {
-    plugins: ["const-enum", "@babel/transform-typescript"],
+    plugins: [
+        "const-enum",
```
oliverlloyd, line 3: I don't think we need this anymore? Not really relevant to emotion so it doesn't need to be changed here but worth being aware of.

src/App.stories.tsx
SiAdcock, line 2: The [Emotion docs imply](https://5faaafd0bd0f3f0008469537--emotion.netlify.app/docs/css-prop#get-started) that if you're using the css prop preset, you shouldn't need to use the JSX pragma as well. Have you found this _not_ to be the case?

package.json
```diff
@@ -46,7 +46,9 @@
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.12.7",
-    "@emotion/babel-preset-css-prop": "^10.0.14",
+    "@emotion/babel-preset-css-prop": "10.0.23",
+    "@emotion/core": "10.0.35",
+    "@emotion/eslint-plugin": "^11.0.0",
```
oliverlloyd, line 51: Is this related to the emotion changes?

package.json
```diff
@@ -46,7 +46,9 @@
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.12.7",
-    "@emotion/babel-preset-css-prop": "^10.0.14",
+    "@emotion/babel-preset-css-prop": "10.0.23",
+    "@emotion/core": "10.0.35",
```
oliverlloyd, line 50: Adding twice because we _really_ want this peer dependency? 😉

.storybook/preview.js
```diff
@@ -25,7 +25,7 @@ const guardianViewports = {
   phablet: {
       name: 'phablet',
       styles: {
-          width: '660px',
+          width: '660px',
```
oliverlloyd, line 28: Was it this change that caused the diffs in Chromatic?

src/components/CommentContainer/CommentContainer.tsx
```diff
@@ -140,10 +142,10 @@ export const CommentContainer = ({
 				onRecommend={onRecommend}
 			/>
 
-			<>
+			<React.Fragment>
```
oliverlloyd, line 145: I'm not following along but is this now how we need to do these fragments now? That would be a shame.

oliverlloyd (commented):
Great to see this working - nice work!

src/components/Avatar/Avatar.stories.tsx
SiAdcock: This is part of the [React Babel Preset](https://babeljs.io/docs/en/babel-preset-react), I would expect it to be already available.

SiAdcock: I suspect that the problem is with the elderly version of `@babel/preset-react` that we're currently using. @liywjl can you try upgrading to v7.9.0 or later?

.storybook/preview.js
liywjl: I have no idea, it seems to have changed focus?? :/ a little concerning

mxdvl:
Closing this as we migrated to Emotion 11 and `@emotion/react` in #454

PR #414, opened 2021-02-02
Update React v17

liywjl:
## What does this change?
Update to React 17 and remove `React` from import scope

## Why?
Emotion 10 uses JSX pragma, and `React` 17 favors JSX over JS. To make Emotion upgrade easier, we should first update `React`

## Changes
- Update React, react types
- Add `@emotion/core` ~(used for JSX)~
- Update `@babel/preset-react`
- ~change `<>` to `<React.Fragment>` as Emotion JSX doesnt pick up fragments: https://github.com/emotion-js/emotion/issues/1706#issuecomment-573317202~
- Remove the need for `import React` (unless `Fragment` used)

https://trello.com/c/Wlpg47zl/2410-update-to-react-17

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "timeago.js": "^4.0.2"
   },
   "peerDependencies": {
-    "@emotion/core": "^10.0.28",
+    "@emotion/core": "10.0.35",
     "@guardian/src-button": "2.7.1",
     "@guardian/src-foundations": "2.7.1",
     "@guardian/src-icons": "2.7.1",
--- a/package.json
+++ b/package.json
@@ -44,9 +44,11 @@
     "@babel/core": "^7.12.10",
     "@babel/plugin-transform-runtime": "^7.9.0",
     "@babel/preset-env": "^7.5.5",
-    "@babel/preset-react": "^7.0.0",
+    "@babel/preset-react": "^7.12.10",
     "@babel/preset-typescript": "^7.12.7",
+    "@emotion/babel-plugin": "^11.1.2",
     "@emotion/babel-preset-css-prop": "^10.0.14",
+    "@emotion/core": "10.0.35",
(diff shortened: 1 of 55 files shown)


package.json
```diff
@@ -152,7 +154,8 @@
     "globals": {
       "ts-jest": {
         "diagnostics": false,
-        "tsConfigFile": "tsconfig.json"
+        "tsConfigFile": "tsconfig.json",
+        "babelConfig": true
```
liywjl, line 158: make sure `ts-jest` is using `babel.config.js` file

shtukas (approved):

buck06191:
I can't be sure but I think that the requirement for explicit `Fragment` statements isn't the case if we can get this all working without the `@jsx` pragma. See https://github.com/guardian/apps-rendering/pull/1005/commits/9a03863d35fdfa8b2c1401b7b383f393987f2d76 where `apps-rendering` removed the `Fragment` import whilst upgrading to React 17 and using Emotion

oliverlloyd:
> I can't be sure but I think that the requirement for explicit `Fragment` statements isn't the case if we can get this all working without the `@jsx` pragma. See [guardian/apps-rendering@9a03863](https://github.com/guardian/apps-rendering/commit/9a03863d35fdfa8b2c1401b7b383f393987f2d76) where `apps-rendering` removed the `Fragment` import whilst upgrading to React 17 and using Emotion

+1 for not having to write `React.Fragment` - `<>` would be a _hard_ habit to break

src/App.tsx
```diff
@@ -398,7 +398,7 @@ export const App = ({
 						)}
 					</div>
 				) : (
-					<React.Fragment>
```
buck06191: Worth updating the PR description to remove references to this being needed once it all goes through

mxdvl:
I think this is superceded by #451

PR #439, opened 2021-02-18
Change urlencoded to json

nitro-marky:
## What does this change?
`'Content-Type': 'application/x-www-form-urlencoded'` to `'Content-Type': 'application/json` as we are posting a JSON

This should allow new users to create an account and post a message.


jfsoul (approved):
👍

PR #447, opened 2021-02-23
Update eslint packages

liywjl:
## What does this change?
- Update linting packages
- supress linting rules (as we had over 250 errors after the update)

## Why?
- The `main` branch linting is broken (we no longer support `react-app` which most of the linting was derived from)
- Linting is out of date compared to DCR and `atoms` repos

## Note
We should look at reverting each linting rule exception in future PRs

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,10 +65,17 @@
     "@types/node": "^14.10.2",
     "@types/react": "^16.9.34",
     "@types/react-dom": "^16.9.5",
+    "@typescript-eslint/eslint-plugin": "^4.15.2",
+    "@typescript-eslint/parser": "^4.15.2",
     "babel-loader": "^8.2.2",
     "babel-plugin-const-enum": "^1.0.1",
     "emotion": "^10.0.27",
-    "eslint-plugin-react-hooks": "^4.0.4",
+    "eslint": "^7.20.0",
+    "eslint-config-airbnb-base": "^14.2.1",
+    "eslint-config-prettier": "^8.0.0",
+    "eslint-plugin-prettier": "^3.3.1",
+    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-react-hooks": "^4.2.0",
     "fetch-mock": "^9.3.1",
     "husky": "^4.2.3",
     "jest": "^26.6.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -107,17 +114,42 @@
...
       "react-hooks"
     ],
+    "extends": [
+      "eslint:recommended",
+      "plugin:prettier/recommended",
+      "plugin:react/recommended",
+      "plugin:@typescript-eslint/recommended"
+    ],
     "rules": {
-      "react-hooks/rules-of-hooks": "error",
```
oliverlloyd, line 115: We probably want to keep this rule

package.json
```diff
@@ -107,17 +114,42 @@
...
-      "react-hooks/exhaustive-deps": "warn"
+      "no-unused-vars": "off",
+      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/no-empty-function": "off",
+      "@typescript-eslint/no-explicit-any": "off",
+      "@typescript-eslint/ban-types": "off",
+      "@typescript-eslint/explicit-module-boundary-types": "off",
+      "@typescript-eslint/no-unused-vars-experimental": "off",
+      "@typescript-eslint/ban-ts-comment": "off",
+      "react-hooks/exhaustive-deps": "off",
```
oliverlloyd, line 142: This was working before, do we need to disable it?

package.json
```diff
@@ -107,17 +114,42 @@
...
+      "prefer-const": "off"
     },
-    "ignorePatterns": [
-      "build"
-    ]
+    "settings": {
+      "react": {
+          "version": "^16.13.1"
+      }
+    }
```
oliverlloyd, line 152: Is this needed for linting?

oliverlloyd (commented):
This brings discussion in line with atoms. But they're both out of sync with DCR.

is there a better centralised linting config we can all align on? There's a lot of pain to big linting refactors so we only want to do them once! What about https://www.npmjs.com/package/@guardian/eslint-config-typescript?

liywjl:
I was having some Linting issues on my machine that were hard to pin down. In the end a fresh install of the repo fixed the issues. 

The undertaking for this is quite large, and we should have better way of aligning linting between each our repos.

I am closing this for now, as it is no longer a blocker

liywjl:
I have created a task for this work https://trello.com/c/qdg7bV4V/2659-align-dcr-atoms-discussion-linting-rules

PR #451, opened 2021-02-26
Add emotion 11

liywjl:
## What does this change?
- Add emotion 11 (`@emotion/react`)
- Use `react-jsx`
- set `jsxImportSource` to `@emotion/react`
- update Source to v3
- Reduce text size as part of update to Source v3
(for text CSS from `textSans.xsmall` to `textSans.xxsmall`; and from `small` to `xsmall` on Button components)

## Why?
I wanted to to just add Emotion 11 (`@emotion/react`) and run everything through Emotion 11 JSX. However Source need to be updated to v3 to support this change.

## Next steps
- remove Emotion 10
- use Emotion 11 css `attr`


shtukas (approved):

package.json
```diff
@@ -29,11 +29,11 @@
   },
   "peerDependencies": {
     "@emotion/core": "^10.0.28",
```
SiAdcock, line 31: Can we delete this now?

package.json
```diff
@@ -29,11 +29,11 @@
...
-    "@guardian/src-icons": "2.7.1",
-    "@guardian/src-link": "2.7.1",
-    "@guardian/src-text-input": "2.7.1",
+    "@guardian/src-button": "^3.3.0",
+    "@guardian/src-foundations": "^3.3.0",
+    "@guardian/src-icons": "^3.3.0",
+    "@guardian/src-link": "^3.3.0",
+    "@guardian/src-text-input": "^3.3.0",
     "@guardian/types": "^3.0.0",
     "emotion": "^10.0.27",
```
SiAdcock, line 38: And this?

package.json
liywjl: good catch, yes this will be turned into `@emotion/react` shortly

package.json
liywjl: yes, we need to migrate each component to @emotion/react before removing (future PR)

PR #454, opened 2021-03-01
Use Emotion 11

liywjl:
## What does this change?
- change use of `emotion` to `@emotion/react`
- remove all uses of `cx` to instead warp styles with `[...]`
- convert `className` to `css`
- remove `@emotion/core` as peer dependency and replace it with `@emotion/react`
- remove `emotion` as dependency

## Why?
We only want 1 version of Emotion in this Repo, so we need to migrate everything over to `@emotion/react`


src/App.test.tsx
```diff
@@ -30,6 +30,9 @@ const aUser = {
 	},
 };
 
+// Extend the timeout, as we are getting failed test because of timeout
+jest.setTimeout(30000);
+
```
oliverlloyd, line 35: 30 seconds seems excessive and suggests there is another issue here, perhaps unrelated to Emotion?

oliverlloyd (approved):
🚀 

+1 for css but we should remove the jest config change from this PR (unless emotion is really causing our test suite to take a long time to run in which case there's a real problem somewhere!)

src/App.test.tsx
liywjl: Its causing CI tests to fail and its kinda outside the scope (likely and API issue causing the slowness?) . I will drop it to a lower number somewhere between 3-5seconds

oliverlloyd: I think if this api call is suddenly taking a long time and it only happens on this branch then we need to understand why so that we don't risk passing through a real issue into production

nitro-marky (approved):
👍

PR #458, opened 2021-03-01
Remove yarn tsc command

liywjl:
## What does this change?
- remove `yarn tsc` command
- `pre-push` uses `yarn build` instead of `yarn tsc`

## Why?
- We want to check that the build works, we should use our command to do that check instead of `yarn tsc` 
- When running `yarn tsc` without options, we would compile `xx.test.tsx` files as well, that in turn would be picked up by Jest

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "pretty-quick --staged && yarn lint && yarn createTsDec && yarn test --watchAll=false && yarn prettier:check"
+      "pre-push": "pretty-quick --staged && yarn lint && yarn build && yarn test --watchAll=false && yarn prettier:check"
     }
   },
   "jest": {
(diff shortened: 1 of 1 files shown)


oliverlloyd:
How can I run `yarn tsc` locally to check I don't have typescript errors?

oliverlloyd:
All the other repos have `yarn tsc` as the standard way to check for issues - I don't think we should diverge from that in discussion. Is there another way to do what you need here?

liywjl:
> How can I run `yarn tsc` locally to check I don't have typescript errors?

you can run `yarn createTsDec`

liywjl:
> All the other repos have `yarn tsc` as the standard way to check for issues - I don't think we should diverge from that in discussion. Is there another way to do what you need here?

You are right, I think maybe I should just remove `createTsDec`, reinstate `yarn tsc`, put the confligs into `tsconfig` and we can they just use that?

liywjl:
closing in favor of https://github.com/guardian/discussion-rendering/pull/459

PR #459, opened 2021-03-01
Remove createTsDec

liywjl:
## What does this change?
- remove `yarn createTsDec` command
- `yarn build` uses `yarn tsc` instead of `yarn createTsDec`
- Move `yarn createTsDec` options to `tsconfig.json`
- Set `emitDeclarationOnly true` option to both `yarn tsc` and rollup to generate types
- Ignore `test` and `stories` when building using TS

## Why?
`yarn createTsDec` was running `yarn tsc` with a bunch of custom options. It makes more sense for those options to be in `tsconfig.json`. It also fixes the issue where we were compiling `xx.test.tsx` files on `pre-push` which would in turn be picked up by Jest

## Why not set `emitDeclarationOnly true` in `tsconfig`?
Storybook seems to be using `tsconfig` and if we set `emitDeclarationOnly` to true, it has issues building. Instead we set it in Rollup and `yarn tsc`


package.json
```diff
@@ -94,7 +94,7 @@
   },
   "scripts": {
     "start": "react-app-rewired start",
-    "build": "rollup --config && yarn createTsDec",
+    "build": "rollup --config && yarn tsc",
```
gtrufitt, line 97: Maybe in this case, it's worth just using the rollup typescript plugin instead?

liywjl: good point

liywjl: I have been unable to get it to work locally :/

gtrufitt: We can update later

shtukas (approved):

liywjl:
will need to test this on a pre-release, as it looking like it might not be bundling all components

mxdvl:
If we use Vite in #609, we can achieve the same thing with [`vite-plugin-dts`](https://www.npmjs.com/package/vite-plugin-dts)

PR #460, opened 2021-03-03
Fix yarn start

liywjl:
## What does this change?
Add `SKIP_PREFLIGHT_CHECK= true` before `react-app-rewired start`

## Why?
`react-app-rewired` isnt actively maintained, and since we have updated other packages, it complains that the dependencies are not the versions it expects. Although this isnt ideal, `babel-loader` has not had a major version jump, and we do not use `react-app-rewired` for jest, so the incompatability is minor. Plus `react-app-rewired` seems to work fine

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,23 +1,33 @@
 {
-    "compilerOptions": {
-        "target": "esnext",
-        "module": "esnext",
-        "lib": ["es2017", "es7", "es6", "dom", "DOM.Iterable"],
-        "outDir": "build",
-        "strict": true,
-        "esModuleInterop": true,
-        "jsx": "react-jsx",
-        "jsxImportSource": "@emotion/react",
-        "moduleResolution": "node",
-        "allowJs": true,
-        "skipLibCheck": true,
-        "allowSyntheticDefaultImports": true,
-        "forceConsistentCasingInFileNames": true,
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -1,23 +1,33 @@
...
+    "jsx": "react-jsx",
+    "jsxImportSource": "@emotion/react",
+    "moduleResolution": "node",
+    "allowJs": true,
+    "skipLibCheck": true,
+    "allowSyntheticDefaultImports": true,
+    "forceConsistentCasingInFileNames": true,
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true
```
oliverlloyd, line 24: Should these be changing?

oliverlloyd (commented):
Do we need `yarn start` to work. I was expecting that we'd eventually get rid of any code relating to running locally

tsconfig.json
liywjl: when running `react-app-rewired` it was keen on formattting this differently

liywjl: looks like it did more than just format :/ good catch

liywjl:
We should instead look at removing `react-app-rewired`

liywjl:
https://github.com/guardian/discussion-rendering/pull/461

PR #477, opened 2021-03-16
Optimistically render first 10 comments and lazily render remaining comments

gtrufitt:
Initial render of comments would be perceived by the user as 3.4x quicker from 1762ms to 512ms on a 4x slow down of CPU.

Screen Shot 2021-03-16 at 17 40 52

(10 refreshes and clicking on the button, x-axis showing value for each run)

### Before - Full render is slow
Screen Shot 2021-03-16 at 15 31 35

### After - Initial rendering early with full rendering not heavily affected
Screen Shot 2021-03-16 at 15 46 28

## Part rendered with loading state

|Before |After |
| ----------- |----------- |
||     |

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@guardian/discussion-rendering",
   "description": "This codebase started as a hack day project by @gtrufitt and @nicl. The purpose is parity of the existing discussion application on Frontend using the discussion API (search for Private Repo).",
-  "version": "7.0.0",
+  "version": "7.1.0-0",
   "author": "",
   "homepage": "https://github.com/guardian/discussion-rendering#readme",
   "license": "Apache",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -243,6 +243,7 @@ export const App = ({
 	);
 	const [isExpanded, setIsExpanded] = useState<boolean>(expanded);
 	const [loading, setLoading] = useState<boolean>(true);
+	const [loadingMore, setLoadingMore] = useState<boolean>(false);
 	const [totalPages, setTotalPages] = useState<number>(0);
 	const [page, setPage] = useState<number>(initialPage || 1);
 	const [picks, setPicks] = useState<CommentType[]>([]);
@@ -273,12 +274,14 @@ export const App = ({
 			// to the total comment amount.
 			// This allows a quick render of minimal comments and then immediately begin rendering
 			// the remaining comments.
+			setLoadingMore(true);
 			const timer = setTimeout(() => {
 				setNumberOfCommentsToShow(comments.length);
+				setLoadingMore(false);
 			}, 0);
 			return () => clearTimeout(timer);
 		}
-	}, [isExpanded]);
+	}, [isExpanded, onExpanded, comments.length]);
 
 	useEffect(() => {
 		setLoading(true);
@@ -567,6 +570,7 @@ export const App = ({
 					))}
 				</ul>
 			)}
+			{loadingMore && <LoadingComments />}
 			{showPagination && (
 				<footer css={footerStyles}>
 					<Pagination
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -251,6 +251,9 @@ export const App = ({
 		setCommentBeingRepliedTo,
 	] = useState<CommentType>();
 	const [comments, setComments] = useState<CommentType[]>([]);
+	const [numberOfCommentsToShow, setNumberOfCommentsToShow] = useState<number>(
+		10,
+	);
 	const [commentCount, setCommentCount] = useState<number>(0);
 	const [mutes, setMutes] = useState<string[]>(readMutes());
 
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -263,7 +267,22 @@ export const App = ({
...
-	}, [isExpanded]);
+
+		if (isExpanded) {
+			// We want react to complete the current work and render, without trying to batch this update
+			// before resetting the number of comments
+			// to the total comment amount.
+			// This allows a quick render of minimal comments and then immediately begin rendering
+			// the remaining comments.
+			setLoadingMore(true);
+			const timer = setTimeout(() => {
```
gtrufitt, line 278: I always feel like this is a code smell but I don't think there's a way to force React not to batch the update (i.e without `setTimeout`, the `setNumberOfCommentsToShow` gets set *before* render, and therefore behaves in the exact same way as before as the state update gets batched)

oliverlloyd: setImmediate`? I agree though, the fact it's needed is hinting at there's a better way but it's not screaming at me. My only thought is to have another component that deals with rendering the list and try to capture the flow inside that using useEffect. I wonder though if a simple setImmediate (or setTimeout) like yuoo've got here isn't simpler?

Not a blocker, just a passing thought, maybe a `useComments` custom hook might work well here to reduce this complexity? Maybe the api could look like:

```typescript
const {
  comments,
  loading,
  loadingMore,
  totalComments,
} = useComments(url, options);
```

src/App.tsx
```diff
@@ -263,7 +267,22 @@ export const App = ({
...
 		}
-	}, [isExpanded]);
+
+		if (isExpanded) {
+			// We want react to complete the current work and render, without trying to batch this update
+			// before resetting the number of comments
+			// to the total comment amount.
+			// This allows a quick render of minimal comments and then immediately begin rendering
+			// the remaining comments.
+			setLoadingMore(true);
```
oliverlloyd, line 277: We could move this to immediatly after we have fetched, where we're setting `loading` to false - might make more sense there?

src/App.tsx
```diff
@@ -263,7 +267,22 @@ export const App = ({
 		) {
 			onExpanded(window.performance.now() - expandedStartTime);
 		}
-	}, [isExpanded]);
+
+		if (isExpanded) {
```
oliverlloyd, line 271: Shall we break this out into it's own `useEffect`? The `onExpanded` timer is something we want to fire at the same point but it's enough of a separate logical action to keep it apart I would say

oliverlloyd (approved):
I think this is brilliant. It's works perfectly, transforms the user experience, and is actually pretty simple code!

There's definately a lot of useEffects here these days and we could do with moving some of this into custom hooks so that we're grouping related useStates and useEffects to make it easier to reason about things but this is just somethiing I wanted to make a note of and not a blocker for here.

package.json
```diff
@@ -6,7 +6,7 @@
   "homepage": "https://github.com/guardian/discussion-rendering#readme",
   "license": "Apache",
   "main": "build/App.js",
-  "module": "src/App.tsx",
+  "module": "build/App.esm.js",
```
liywjl, line 9: feels like this is outside the scope of this PR

rollup.config.js
```diff
@@ -33,11 +33,11 @@ module.exports = {
         clear({
             targets: ['build/'],
         }),
+        commonjs(),
```
liywjl, line 36: same here, I assume there is a deeper reason for this change that would get lost in PRs

liywjl (approved):
Echoing the same points as @oliverlloyd , this PR is great 👍 

You can maybe also remove `jest.setTimeout(20000);` from`src/App.test.tsx`

rollup.config.js
gtrufitt: Yup split out into https://github.com/guardian/discussion-rendering/pull/478

src/App.tsx
gtrufitt: Aye, totally, I had that originally but got your feeling of "There's definately a lot of useEffects" but makes more sense seperate.

src/App.tsx
JamieB-gu: Just to check my understanding, this is to defer execution?

JamieB-gu (commented):
Nice! 🔥 📈

src/App.tsx
gtrufitt: > setImmediate

Only IE supports, but essentially, yea.

> I wonder though if a simple setImmediate (or setTimeout) like yuoo've got here isn't simpler?

Yeah, I feel like it. Took me a bit to walk through this file already, it's possibly worth refactoring if we implement similar elsewhere but for now seems simple enough.

> maybe a useComments custom hook might work well here to reduce this complexity

Yeah, that would help make App.tsx more manageable. I'll give it a go and follow up with another PR.

gtrufitt: > Just to check my understanding, this is to defer execution?

Yes, it pushes the execution of the callback to the back of the event queue, so React will complete it's update before executing it, and therefore not batch this `setState` update before render.

PR #498, opened 2021-06-01
Add prop & styles for full width styling

OllysCoding:
## What does this change?

Creates an optional prop which adds `width: 100%` to the comments container.

## Why?

On articles where no comments span the full width on the available container, the comments section shrinks to the size of the largest comment

image

https://trello.com/c/xurRBeNp/235-comment-section-width-not-correct

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,6 +31,7 @@ type Props = {
 	baseUrl: string;
 	pillar: Theme;
 	isClosedForComments: boolean;
+	isFullWidth?: boolean;
 	commentToScrollTo?: number;
 	initialPage?: number;
 	pageSizeOverride?: PageSizeType;
@@ -68,6 +69,10 @@ const commentContainerStyles = css`
 	list-style-type: none;
 	padding-left: 0;
 	margin: 0;
+`;
+
+const fullWidthStyles = css`
+	width: 100%;
 `;
 
 const picksWrapper = css`
@@ -218,6 +223,7 @@ export const App = ({
 	shortUrl,
 	pillar,
 	isClosedForComments,
+	isFullWidth,
 	initialPage,
 	commentToScrollTo,
 	pageSizeOverride,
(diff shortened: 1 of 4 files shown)


shtukas (approved):

src/App.tsx
```diff
@@ -31,6 +31,7 @@ type Props = {
 	baseUrl: string;
 	pillar: Theme;
 	isClosedForComments: boolean;
+	isFullWidth?: boolean;
```
liywjl, line 34: where does this prop come from?

liywjl (commented):
Would be great to maybe add a storybook story to highlight these issues. It has appeared several times and would be great to have a visual regression test to catch future cases

OllysCoding:
> Would be great to maybe add a storybook story to highlight these issues.

Will get on this!

src/App.tsx
OllysCoding: It would be passed in from DCR or any other consuming frontend, for example we'd add the prop on DCR [here](https://github.com/guardian/dotcom-rendering/blob/b3211ac220a7a7f82dff7d44a4f1b5813127d19b/src/web/components/Discussion.tsx)

liywjl: would we ever want `discussion-rendering` to not be full width?

liywjl:
I just realised this was running on `discussion-rendering` and not DCR. I think it might be worth adding a storybook on DCR about just discussion and its wrapper, as it is not always obvious how they interact with one another

OllysCoding:
> I think it might be worth adding a storybook on DCR about just discussion and its wrapper

Can you expand on this a little bit, not quite following sorry!

src/App.tsx
OllysCoding: I wasn't 100% sure on this, but knew it was being used other frontend services as well, and wasn't sure how their implementation & containers would work or if they would place nice with it. The prop is mostly just a safety guard for that reason, but it could be removed and leave it up to someone checking when they update

OllysCoding:
Closing as I've found that this issue can be resolved in DCR only. It doesn't appear to be a bug within `discussion-rendering` at all.

PR #524, opened 2021-09-24
Add contributor badge

DavidLawes:
## What does this change?
Via modtools a Guardian-employee profile can be marked as a contributor (eg a freelancer). 

In frontend we marked comments from contributors with an associated badge. https://github.com/guardian/frontend/blob/main/discussion/app/views/fragments/commentBadges.scala.html#L8

This change adds the equivalent functionality. Note that frontend assumes we show either Staff or Contributor badges, not both.

Liveblogs haven't been migrated to dcr yet and contributor badges in liveblogs are showing up as expected, eg https://www.theguardian.com/sport/live/2021/sep/06/county-cricket-warwickshire-v-hampshire-yorkshire-v-somerset-and-more-live#comment-151643779

Currently, for articles rendered via dcr, contributors did not have a badge:
Screenshot 2021-09-24 at 17 13 38

After this change, contributors have a badge:
Screenshot 2021-09-24 at 17 14 26

## Why?
This issue has been reported by CP & Moderators.

https://trello.com/c/J83m02bt

### DIFF
--- a/src/components/Badges/Badges.tsx
+++ b/src/components/Badges/Badges.tsx
@@ -34,7 +34,7 @@ const guardianPickLabel = css`
 	color: ${neutral[7]};
 `;
 
-export const GuardianStaff = () => (
+export const GuardianStaff = ({ role }: { role: 'Staff' | 'Contributor' }) => (
 	<div css={staffBadge}>
 		<svg
 			width="36"
@@ -48,7 +48,7 @@ export const GuardianStaff = () => (
 				d="M21.2 4.4c2.3.4 5.3 2 6.3 3.1v5.2H27L21.2 5v-.6zm-2.2.4c-4 0-6.3 5.6-6.3 13.2 0 7.7 2.2 13.3 6.3 13.3v.6c-6 .4-14.4-4.2-14-13.8A13.3 13.3 0 0 1 19 4v.7zm10.4 14.4l-1.9.9v8.6c-1 1-3.8 2.6-6.3 3.1V19.9l-2.2-.7v-.6h10.4v.6z"
 			></path>
 		</svg>
-		<p css={[labelText, staffLabel]}>Staff</p>
+		<p css={[labelText, staffLabel]}>{role}</p>
 	</div>
 );
 
--- a/src/components/Badges/Badges.tsx
+++ b/src/components/Badges/Badges.tsx
@@ -34,7 +34,7 @@ const guardianPickLabel = css`
 	color: ${neutral[7]};
 `;
 
-export const GuardianStaff = ({ role }: { role: 'Staff' | 'Contributor' }) => (
+export const GuardianStaff = () => (
 	<div css={staffBadge}>
 		<svg
 			width="36"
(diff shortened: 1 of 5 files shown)


DavidLawes:
[Contributor commenter badges not showing](https://trello.com/c/J83m02bt/370-contributor-commenter-badges-not-showing)

src/components/Comment/Comment.tsx
```diff
@@ -328,12 +328,17 @@ export const Comment = ({
...
 
-	const showStaffbadge = comment.userProfile.badge.find(
+	const showStaffBadge = comment.userProfile.badge.find(
 		(obj) => obj['name'] === 'Staff',
 	);
 
 	const showPickBadge = comment.status !== 'blocked' && isHighlighted;
 
+	// A contributor could be e.g. a freelancer that commonly comments on articles.
+	// In frontend we check/display the Staff badge else we check/display the
```
oliverlloyd, line 338: If this Frontend code is relevant here should we link to it?

src/components/Comment/Comment.tsx
```diff
@@ -455,9 +460,11 @@ export const Comment = ({
 										</div>
 									</Row>
 									<Row>
-										{showStaffbadge ? (
+										{showStaffBadge || showContributorBadge ? (
```
oliverlloyd, line 463: This feels a bit like these two bits of logic are fighting each other. Maybe we can break this apart and simply have

```
{showStaffBadge && <GuardianStaff role='Staff' />}
{showContributorBadge && <GuardianStaff role='Contributor' />}
```

?

If we explicitly want to prevent both badges showing at the same time you could


```
{showStaffBadge && <GuardianStaff role='Staff' />}
{showContributorBadge && !showStaffBadge && <GuardianStaff role='Contributor' />}
```

not sure if this extra check is needed or not. Ideally the server would decide this type of thing and the rendering layer just does what it is asked to do

src/components/Badges/Badges.tsx
```diff
@@ -34,7 +34,7 @@ const guardianPickLabel = css`
 	color: ${neutral[7]};
 `;
 
-export const GuardianStaff = () => (
+export const GuardianStaff = ({ role }: { role: 'Staff' | 'Contributor' }) => (
```
oliverlloyd, line 37: You could also have a second component, like  `GuardianContributor` and then there'd be no need for the prop? I think if I were starting from scratch and these were the requirements from day one I would have done that

Fully admit that this comment is a bit opinionated and feely though so whatever feels best to you

oliverlloyd (approved):
Nice! Always happy to see Discussion getting some ❤️ !

Just a few comments, nothing earth shattering and I'm happy to leave them with you to do as you see fit. On thing though, it'd be ace to get a screenshot in the description showing how this new badge looks 🙏  Thanks!

src/components/Badges/Badges.tsx
DavidLawes: Actually, this was my initial thought too :) Looking at how the Contributor badge renders via frontend it's identical to the Staff badge, so the GuardianContributor component was a duplication with just the inner text differing. So, I tried to be tidy by giving the component a `role` - might regret that choice soon!

DavidLawes:
> Nice! Always happy to see Discussion getting some ❤️ !
> 
> Just a few comments, nothing earth shattering and I'm happy to leave them with you to do as you see fit. On thing though, it'd be ace to get a screenshot in the description showing how this new badge looks 🙏 Thanks!

Thanks - have just figured out how to test this all locally and I need to make some changes. When I'm a bit happier with the logic I'll update this PR to include some screenshots :)

src/components/Badges/Badges.tsx
oliverlloyd: Ha, well. One reason for not using `role` is that this name is used _all over_ the place in Guardian code. Around images mainly. Incidentally, the same thing is also called `weight` sometimes but that's just to keep us paying attention.

PR #532, opened 2021-11-09
migrate discussion-rendering to source v4

jamie-lynch:
## Why?

We have published a Source v4 release candidate, and are keen for dotcom-rendering and apps-rendering to be first benefit from the improved developer ergonomics 🎉   In order to do this, we must update packages that they depend on, so as to avoid bundling both v3 and v4 of Source.

We have written an eslint plugin that has automated the majority of the upgrade. 

## What has changed?

Source now comprises 3 packages:

- `@guardian/source-foundations`: equivalent to `@guardian/src-foundations`
- `@guardian/source-react-components`: contains all components from packages that were prefixed `@guardian/src-*`
- `@guardian/source-react-components-development-kitchen`: New and experimental components

## Workarounds

There are a few places that use the old `palette` object to do `palette[pillarToString(pillar)][400]`.

As `palette` has now been deprecated in favour of importing the component parts directly, this is no longer possible. To replicate the existing behaviour in a type safe way, whilst making minimal changes, the required `palette` object (containing only the parts required here, rather than the full object available previously) has been created in `src/lib/palette.ts`. In a future PR, this pattern could well be refactored to provide a nicer solution, potentially combining it with the `pillarToString` function.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,5 +1,5 @@
 import React, { useEffect } from "react"
-import { FocusStyleManager } from "@guardian/src-foundations/utils"
+import { FocusStyleManager } from "@guardian/source-foundations"
 import MockDate from 'mockdate';
 import { mockFetchCalls } from '../src/lib/mockFetchCalls';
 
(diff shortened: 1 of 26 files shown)


lucymonie (commented):
The palette workaround seems fine though it certainly highlights a bit of an issue with colour selection here.

OllysCoding (approved):

PR #549, opened 2022-01-18
Accessibility improvements

oliverlloyd:
## What does this change?
This PR adds 2 accessibility improvements

1) An aria label for the Recommend button
2) It removes `FocusLock`, improving keyboard access to the filters

## Why?
We don't need to lock focus on these filters because it's logical to tab through them and the previous package we were using was causing focus to stop on two hidden divs, which would be confusing for people who use screen readers

## Focus before
It's hard to see from the recording but focus is landing on two hidden divs when tabbing through the menus

https://user-images.githubusercontent.com/1336821/149975274-a0a52550-ab6a-48d5-9082-77526d8e8fc4.mov

## Focus after

https://user-images.githubusercontent.com/1336821/149975049-5b0d4200-c7e4-4bba-88aa-1ea3c50cb1e5.mov

## Wait, look, it's James!
@jamesgorrie - you left but we never stopped staring at your sunflower avatar

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
   "dependencies": {
     "babel-plugin-emotion": "^10.0.33",
     "customize-cra": "^1.0.0",
-    "react-focus-lock": "^2.2.1",
     "timeago.js": "^4.0.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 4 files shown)


OllysCoding (commented):
LGTM! Nice shoutout to @jamesgorrie as well!

OllysCoding (approved):

PR #550, opened 2022-01-21
fix: Re-add anchor icon to comment permalink.

AshCorr:
## What does this change?

Shows the permalink icon when the comment is hovered over by the user. This used to be the old behaviour that was used by Frontend

## Frontend version
image
image

|                                                      Before                                                     |                                                      After                                                      |
|:---------------------------------------------------------------------------------------------------------------:|:---------------------------------------------------------------------------------------------------------------:|
| image | image |
| image | image |

### DIFF
--- a/src/components/Timestamp/Timestamp.tsx
+++ b/src/components/Timestamp/Timestamp.tsx
@@ -28,6 +28,16 @@ const timeStyles = css`
 	white-space: nowrap;
 `;
 
+const anchorStyles = css`
+	display: none;
+	width: 15px;
+	height: 15px;
+	vertical-align: middle;
+	margin-left: 0.125rem;
+	max-height: 0.9375rem;
+	fill: #ffffff;
+`;
+
 export const Timestamp = ({
 	isoDateTime,
 	webUrl,
@@ -54,6 +64,15 @@ export const Timestamp = ({
 			<time dateTime={isoDateTime.toString()} css={timeStyles}>
 				{timeAgo}
 			</time>
+
+			<span css={anchorStyles} className="comment-permalink-icon">
+				<svg viewBox="0 0 15 15" xmlns="http://www.w3.org/2000/svg">
+					<path
+						fill="#898989"
+						d="M5.725,13.348c-0.979,1.021-2.6,1.055-3.623,0.076c-1.02-0.979-1.055-2.602-0.076-3.623l8.424-8.79 c0.732-0.766,1.95-0.792,2.717-0.057c0.768,0.734,0.791,1.949,0.059,2.715l-6.894,7.193C5.84,11.375,5.03,11.393,4.52,10.9 c-0.511-0.488-0.528-1.299-0.04-1.81l5.324-5.553l0.337,0.321L4.816,9.413c-0.311,0.325-0.3,0.84,0.025,1.152 c0.326,0.312,0.841,0.301,1.15-0.024l6.896-7.193c0.557-0.581,0.536-1.501-0.043-2.058c-0.581-0.557-1.504-0.537-2.058,0.043 l-8.427,8.79c-0.798,0.837-0.771,2.163,0.063,2.965c0.836,0.802,2.164,0.772,2.965-0.062l6.853-7.152l0.338,0.323L5.725,13.348z"
+					/>
+				</svg>
+			</span>
 		</a>
 	);
 };
--- a/src/components/Comment/Comment.tsx
+++ b/src/components/Comment/Comment.tsx
@@ -131,6 +131,10 @@ const commentWrapper = css`
 	border-top: 1px solid ${border.secondary};
 	display: flex;
 	padding: ${space[2]}px 0;
+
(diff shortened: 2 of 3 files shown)


src/components/Timestamp/Timestamp.tsx
```diff
@@ -54,6 +64,15 @@ export const Timestamp = ({
 			<time dateTime={isoDateTime.toString()} css={timeStyles}>
 				{timeAgo}
 			</time>
+
+			<span css={anchorStyles} className="comment-permalink-icon">
+				<svg viewBox="0 0 15 15" xmlns="http://www.w3.org/2000/svg">
```
AshCorr, line 69: I debated on adding this SVG to source but after comparing it to the existing SVG's source has I noticed that the styling is completely different.

src/components/Timestamp/Timestamp.tsx
```diff
@@ -54,6 +64,15 @@ export const Timestamp = ({
 			<time dateTime={isoDateTime.toString()} css={timeStyles}>
 				{timeAgo}
 			</time>
+
+			<span css={anchorStyles} className="comment-permalink-icon">
```
AshCorr, line 68: Not quite sure if theres a way to apply the hover behaviour without using a className unless I extract the css styling out of this class.

AshCorr:
[Discussion Rendering: The permalinks are missing the little 'link' icon next to the timestamp](https://trello.com/c/UymZBOLC/567-discussion-rendering-the-permalinks-are-missing-the-little-link-icon-next-to-the-timestamp)

src/components/Timestamp/Timestamp.tsx
```diff
@@ -28,6 +28,16 @@ const timeStyles = css`
...
 
+const anchorStyles = css`
+	display: none;
+	width: 15px;
+	height: 15px;
+	vertical-align: middle;
+	margin-left: 0.125rem;
+	max-height: 0.9375rem;
+	fill: #ffffff;
+`;
```
jamesgorrie, line 39:
```suggestion
const anchorStyles = css`
	display: none;
	width: 15px;
	height: 15px;
	vertical-align: middle;
	margin-left: 0.125rem;
	max-height: 0.9375rem;
	fill: #ffffff;
	&:hover {
		display: inline-block;
	}
`;
```

Would this work in response to [your comment](https://github.com/guardian/discussion-rendering/pull/550/files#r789783418)?

This might might negate the need to add the CSS in [two places](https://github.com/guardian/discussion-rendering/pull/550/files#diff-726f6a4e18a031a44d7ef6e6019e0bbf742d7755f4015edf9c6e955839efdefaR160).

src/components/Timestamp/Timestamp.tsx
jamesgorrie: If this is the paperclip icon, I am still in awe that we use it. I can't see any `link` icon [in source](https://guardian.github.io/source/?path=/docs/packages-source-react-components-icons--playground) - I wonder if it's time to ask if we could have it?

jamesgorrie (approved):

src/components/Timestamp/Timestamp.tsx
AshCorr: Gave it a go, afraid it didn't work. The problem is the link icon needs to be shown when any part of the comment is hovered over - not just the timestamp - which is why I needed to apply some CSS styling in the parent Comment class to capture that hover

src/components/Timestamp/Timestamp.tsx
OllysCoding: For adding SVG's that aren't in source, we can add them into `dotcom-rendering/src/static/icons`. 

Then you can import and use them like: 

```jsx

import MyIcon from '@root/src/static/icons/my-icon.svg'

...

<MyIcon />
```

I think this is the preferred way of adding SVG's in DCR

OllysCoding: ^^ Please ignore the above I mistakenly thought this was a DCR pull request ...

src/components/Timestamp/Timestamp.tsx
OllysCoding: I think the way to do this would be 

```css
&::hover {
    display: inline-block;
}```

OllysCoding: Generally we try to avoid using any class names so hopefully we can figure out a way for this!

src/components/Timestamp/Timestamp.tsx
jamesgorrie: Definitely not a blocker either, just wondered if we could work with them on getting something that actually indicates what it is.

src/components/Timestamp/Timestamp.tsx
jamesgorrie: [Potential fix](https://github.com/emotion-js/emotion/issues/1771)?

AshCorr:
Won't be re-implemented in DCR

PR #559, opened 2022-03-22
Refactor expansion

oliverlloyd:
## What does this change?
This PR simplifies how expansion is handled.

It removes the callback we were using to track how long it took to expand as this were adding complexity to the code and was no longer needed or used. We previously used this callback as part of the migration to Emotion 11 work.

It adds a step to ensure that the `isExpanded` state is set whenever the `expanded` prop changes. There are use cases where the `Discussion` component in the parent is reloaded with a new value for this prop (permalinks) and we need to respect this.

This PR closes https://github.com/guardian/dotcom-rendering/issues/4297


mxdvl (approved):
neat!

PR #568, opened 2022-04-19
Remove expand button

oliverlloyd:
## What does this change?
This removes the `View more comment` expand button and associated tests

## Why?
We're going to move this button to the calling container so that we can fix its height and reduce layout shifts

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -395,7 +395,9 @@ export const App = ({
 
 	const showPagination = totalPages > 1;
 
-	if (!isExpanded && loading) return <></>;
+	if (!isExpanded && loading) {
+		return <span data-testid="loading-comments"></span>;
+	}
 
 	if (!isExpanded) {
 		return (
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -351,6 +351,7 @@ export const App = ({
 		rememberFilters(newFilterObject);
 		// Filters also show when the view is not expanded but we want to expand when they're changed
 		setIsExpanded(true);
+		if (typeof onExpand === 'function') onExpand();
```
mxdvl, line 354: If you feel it…

```suggestion
		onExpand?.();
```

mxdvl (approved):
Nice simplification.

Any way we could get an official release of `discussion-rendering`?

oliverlloyd:
> Any way we could get an official release of `discussion-rendering`?

Absolutely. I was waiting on the merge to `main` before releasing the proper version

PR #571, opened 2022-04-29
refactor: resolve git.io link

mxdvl:
## What does this change?

Resolve the `git.io` links.

## Why?

https://github.blog/changelog/2022-04-25-git-io-deprecation/


bryophyta:
These currently seem to resolve to slightly different URLs (different snapshots of `main`?) but the code in both seems to be the same, so looks good to me 👍

bryophyta (approved):

PR #572, opened 2022-05-09
fix: add CSS resets

mxdvl:
## What does this change?

Adds [Source’s reset CSS](https://github.com/guardian/source/blob/main/packages/@guardian/source-foundations/src/utils/resets.stories.mdx)

## Why?

These are currently missing from Storybook, which prevents us from having Chromatic screenshots that look like production

## Screenshots

First comment


bryophyta (approved):
This looks good to me 👍

PR #575, opened 2022-05-16
fix(np): run np from package

mxdvl:
## What does this change?

Allow releasing without having installed `np` globally.

## Why?

It’s already in the package deps. We tested it locally with @bryophyta


bryophyta (approved):
+1

PR #587, opened 2022-07-26
Bump version to 10.1.2

georgeblahblah:
## What does this change?

- Bumps `package.json` version to `10.1.2` ahead of publishing this package to `npm`

## Why?

- To reflect a8655b3028f654d7ff700903011f2b3e4f537941 - separating hours & minutes with a dot (`.`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@guardian/discussion-rendering",
   "description": "This codebase started as a hack day project by @gtrufitt and @nicl. The purpose is parity of the existing discussion application on Frontend using the discussion API (search for Private Repo).",
-  "version": "10.1.1",
+  "version": "10.1.2",
   "author": "",
   "homepage": "https://github.com/guardian/discussion-rendering#readme",
   "license": "Apache",


mxdvl (commented):
This is so manual… can we set up automated publishing?

cc. @bryophyta

georgeblahblah:
> This is so manual… can we set up automated publishing?
> 
> cc. @bryophyta

@mxdvl I agree, an automated process would definitely reduce the amount of work required. There was a [small discussion](https://chat.google.com/room/AAAAag0I08g/S1-n-wAmQKc) where we decided to postpone automating it for now. I don't know much about how to set up automated publishing but I'd be very happy to pair on it!

bryophyta:
Sounds good! iirc we'd been planning to pair on adding `changesets` when we had time, right Max? I'll message you both off this thread and see  if we can get something organised for this week

oliverlloyd:
I think you don't need this PR, you can just `yarn release` and the `np` package takes care of this?

But yes, changesets sound good 👍

PR #589, opened 2022-07-29
Show the second comment box less

### LINKED ISSUE #588: Two comment boxes appear in view at the same time on short discussions
We render the `CommentForm` both at the top and the bottom of a discussion but in some cases, where there are very few comments, this can lead to the same form being shown twice in the viewport

Image

See original email

> Hi there,
>
> I've just opened discussion here in premod and noticed that there are two comment boxes at the bottom of the page. Chrome and Safari both display the same issue.
>
> https://www.theguardian.com/lifeandstyle/2022/jul/29/my-husband-is-perpetually-grumpy-and-im-desperate-for-some-light-relief-what-can-i-do
>
> Cheers
> Keiran
> --
> -----
>
> Keiran McPhillips
>
> Community Moderator
>
> The Guardian | Australia

oliverlloyd:
## What does this change?
We now only show the second comment box when there are more than 10 comments. This closes #588 

## Why?
Having two comment boxes so close together on the page makes no sense, we only need to start showing the second one after the discussion starts to get longer

| Before | After |
| -------- | ------- |
| Screenshot 2022-07-29 at 10 10 58 | Screenshot 2022-07-29 at 10 04 41 |

### DIFF
--- a/src/App.stories.tsx
+++ b/src/App.stories.tsx
@@ -154,6 +154,34 @@ export const LoggedIn = () => (
 LoggedIn.story = {
 	name: 'when logged in and expanded',
 };
+
+export const LoggedInShortDiscussion = () => (
+	<div
+		css={css`
+			width: 100%;
+			max-width: 620px;
+		`}
+	>
+		<App
+			shortUrl="p/39f5a" // Two comments"
+			pillar={ArticlePillar.News}
+			isClosedForComments={false}
+			user={aUser}
+			baseUrl="https://discussion.theguardian.com/discussion-api"
+			additionalHeaders={{
+				'D2-X-UID': 'testD2Header',
+				'GU-Client': 'testClientHeader',
+			}}
+			expanded={true}
+			onPermalinkClick={() => {}}
+			apiKey=""
+		/>
+	</div>
+);
+LoggedInShortDiscussion.story = {
+	name: 'when logged in but only wo comments made',
+};
+
 export const LoggedOutHiddenNoPicks = () => (
 	<div
 		css={css`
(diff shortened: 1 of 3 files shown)


src/App.stories.tsx
```diff
@@ -128,6 +128,60 @@ LoggedInHiddenNoPicks.story = {
...
+			}}
+			expanded={true}
+			onPermalinkClick={() => {}}
+			apiKey=""
+		/>
+	</div>
+);
+LoggedInShortDiscussion.story = {
+	name: 'when logged in but only wo comments made',
+};
```
jamesgorrie, line 183: Is the reason to move it to a story that if someone changes the logic, it will show up in the visual diff, and someone would approve the change as intentional?

jamesgorrie (approved):
Nice and elegant. Is there a reason we went for 10 (I just wonder if that's quite a lot of space)?

src/App.stories.tsx
oliverlloyd: No, it was because the test wasn't rendering any comments 😐 so I wasn't able to verify the logic I was adding. I figured if the storybook tests are working it makes sense to favour using them rather than trying to support two types of test doing the same thing

oliverlloyd:
> Is there a reason we went for 10 (I just wonder if that's quite a lot of space)?

It's arbitrary. Comments can vary quite a lot in terms of how long they are so there's not really a way to pick based on expected height so I went with 10 because a) it's probably a fair indication that the discussion is getting traction and more comments will follow soon and b) because it's a nice round number

PR #595, opened 2022-09-07
Mxdvl/package fixes

mxdvl:
## What does this change?

Reorganise packages, bump versions and remove `timeago.js`.

## Why?

So they are not passed down to consumers if they are only development dependencies.
The package `timeago.js` is redundant with our implementation in `@guardian/libs`

Flagged by @georgeblahblah in https://github.com/guardian/dotcom-rendering/pull/5949#discussion_r963852620


mxdvl:
This is an absolute mess.

mxdvl:
Split into:
- #596 
- #597
- #598

PR #598, opened 2022-09-07
Align peer and dev dependencies

mxdvl:
## What does this change?

Align peer and dev dependencies and run a script to make sure it stays that way.

## Why?

They were currently out of sync.


georgeblahblah (commented):
Looks good! Do you think we need a changeset given the changes to the `peerDependencies` versions?

mxdvl:
This is dependant on #600

georgeblahblah (approved):

PR #599, opened 2022-09-08
chore: remove vscode workbench colour settings

georgeblahblah:
## What does this change?

- Removes workbench colour definitions from VSCode settings

## Why?

- I'm not sure these should be configured by the repository

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,9 +15,5 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   },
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
These were added as a way to give visual clues when working accross various `*-rendering` repos in #402.

I do not see much value in it, so I’m happy with this change. Might be worth checking with @oliverlloyd ?

oliverlloyd (approved):
I agree. These were designed to help when switching between repos but our workflows are different these days

PR #600, opened 2022-09-08
refactor(timeAgo): use the @guardian/libs one

mxdvl:
## What does this change?

Remove `timeago.js` in favour of our own `@guardian/libs` one.

## Why?

We have a similar abstraction in [`@guardian/libs/timeAgo`](https://github.com/guardian/csnx/blob/main/libs/%40guardian/libs/src/datetime/timeAgo.ts)

## Screenshots

image

### DIFF
--- a/src/components/Timestamp/Timestamp.tsx
+++ b/src/components/Timestamp/Timestamp.tsx
@@ -34,7 +34,7 @@ export const Timestamp = ({
 	commentId,
 	onPermalinkClick,
 }: Props) => {
-	let [timeAgo, setTimeAgo] = useState(dateFormatter(isoDateTime));
+	const [timeAgo, setTimeAgo] = useState(dateFormatter(isoDateTime));
 
 	useInterval(() => {
 		setTimeAgo(dateFormatter(isoDateTime));
--- a/.changeset/calm-eggs-laugh.md
+++ b/.changeset/calm-eggs-laugh.md
@@ -1,0 +1,5 @@
+---
+'@guardian/discussion-rendering': patch
+---
+
+Usage of `timeAgo` from @guardian/libs. The timestamps no longer have hours and minutes if they happened longer than 48h ago.
--- a/package.json
+++ b/package.json
@@ -20,9 +20,6 @@
   "files": [
     "build/**/*"
   ],
-  "dependencies": {
-    "timeago.js": "^4.0.2"
-  },
   "peerDependencies": {
     "@emotion/react": "^11.1.5",
     "@guardian/libs": "^3.1.0",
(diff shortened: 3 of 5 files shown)


src/components/Timestamp/Timestamp.tsx
```diff
@@ -28,16 +28,20 @@ const timeStyles = css`
...
 	commentId,
 	onPermalinkClick,
 }: Props) => {
-	let [timeAgo, setTimeAgo] = useState(dateFormatter(isoDateTime));
+	const epoch = Date.parse(isoDateTime);
+	const [timeAgo, setTimeAgo] = useState(dateFormatter(epoch));
 
 	useInterval(() => {
-		setTimeAgo(dateFormatter(isoDateTime));
+		setTimeAgo(dateFormatter(epoch));
```
oliverlloyd, line 44:
```suggestion
		timeAgoLib(epoch, { verbose: true, daysUntilAbsolute: 1 });
```

Might as well inline?

src/components/Timestamp/Timestamp.tsx
```diff
@@ -28,16 +28,20 @@ const timeStyles = css`
...
+
 export const Timestamp = ({
 	isoDateTime,
 	webUrl,
 	commentId,
 	onPermalinkClick,
 }: Props) => {
-	let [timeAgo, setTimeAgo] = useState(dateFormatter(isoDateTime));
+	const epoch = Date.parse(isoDateTime);
+	const [timeAgo, setTimeAgo] = useState(dateFormatter(epoch));
```
oliverlloyd, line 41: I'm not strong on this but it maybe we could rename this state to something else and then we don't need to rename `timeAgo` to `timeAgoLib`?

oliverlloyd (approved):
Just a couple of non blocking comments/thoughts, I'm happy for you to ignore or action as you feel inclined

More power to Guardian time ✊ !

src/components/Timestamp/Timestamp.tsx
mxdvl: You have to inline twice and we don’t have any testing of the `setInterval` working as expected, so I opted for abstracting that away.

src/components/Timestamp/Timestamp.tsx
mxdvl: Currently the two uses of `timeAgo` do not clash, so I’ve opted for not renaming the import.

src/components/Timestamp/Timestamp.tsx
mxdvl: I actually refactored to keep the changes inside the `dateFormatter`, which means we can keep this a patch change.

PR #601, opened 2022-09-08
Unify Jest version and use bundled `jsdom`

mxdvl:
## What does this change?

Bump Jest to a version that includes JSDOM.

## Why?

Simplify dependency graph.


georgeblahblah (approved):
I'm curious about these "new" lint warnings - I'm not sure I've seen them before, but I'm not sure why they'd show up now!

mxdvl:
> I'm curious about these "new" lint warnings - I'm not sure I've seen them before, but I'm not sure why they'd show up now!

These are due to ESLint going from v7 to v8. There’s a few differences that now come standard.

mxdvl:
I have no idea why the actions are failing… seems to be an issue with `ts-loader`.

PR #609, opened 2022-10-21
Use Vite

mxdvl:
## What does this change?

Use Vite for the test page & Storybook

## Why?

Makes things faster and simpler. 

Hopefully this will help the dependabot PRs to pass! We also want to bump Source to the latest version.


AshCorr (approved):
Don't know too much about Vite and its configuration but I also don't to gatekeep this PR!

As far as I understand this should affect how DCR uses discussion-rendering right? Just the bundling process for Storybook? With that in mind I'd quite like to see how it goes!

.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<script>
+    window.global = window;
```
OllysCoding, line 2: What does this do?

src/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
OllysCoding, line 1: What does this mean?

.storybook/preview-head.html
mxdvl: This is needed for some of the Storybook addons and the fix is suggested in https://github.com/storybookjs/storybook/issues/18399 and https://github.com/storybookjs/builder-vite/blob/main/packages/builder-vite/README.md#migration-from-webpack--cra

OllysCoding (approved):
+1 A few questions to benefit my own understanding, but this looks good!

src/vite-env.d.ts
mxdvl: This enables [IntelliSense for TypeScript](https://vitejs.dev/guide/env-and-mode.html#intellisense-for-typescript)

vite.config.ts
```diff
@@ -0,0 +1,37 @@
...
+        jsxImportSource: "@emotion/react",
+    }),
+  ],
+  build: {
+    outDir: "build",
+    lib: {
+      entry: "./src/App.tsx",
+      formats: ["cjs","es"],
+      fileName: (format) => {
+        const path = format === "es" ? pkg.module : pkg.main
```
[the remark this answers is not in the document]
mxdvl: This ensures that the file name is coming from the `package.json`, rather than having a loose contract between it and the build config.

Added a comment in 9679bb7f608e993d1136504c96d685ce6531fb57.

PR #610, opened 2022-10-21
Use @guardian/eslint

mxdvl:
## What does this change?

Use the `@guardian/eslint-typescript-config` package

## Why?

Consistency


mxdvl:
This is not a priority

PR #611, opened 2022-10-24
Fix build process

mxdvl:
## What does this change?

Imports some packages explicitely.

## Why?

The current build is failing. It was probably working before because `yarn` and `npm` shamefully hoist modules.

#609 depends on this of the compress size action to work. So do most dependabot PRs.

Addresses https://github.com/guardian/dotcom-rendering/issues/6264


AshCorr (approved):
👍

PR #632, opened 2022-10-28
Fix emotion `css`

mxdvl:
## What does this change?

Adjust the vite configuration to use Emotion’s JSX by adding the correct babel and esbuild options. Comments have been added to give as detailed an explaination as I could articulate.

Do some housekeeping:
- refactor: remove React imports
- fix: catch formatting errors with prettier
- feat: add Babel options types
- feat: format vite config
- chore(deps): bump Emotion packages

## Why?

Version 11.0.1 fails to produce working class names from Emotion’s `css` tagged templates.


vite.config.ts
```diff
@@ -2,17 +2,28 @@ import { defineConfig } from 'vite';
...
 import pkg from './package.json';
+import { Plugin } from 'vite';
 
 /** Get build directory from package.json */
 const outDir = pkg.module.split('/')[0];
 
 // https://vitejs.dev/config/
 export default defineConfig({
+	esbuild: {
+		// These values
```
bryophyta, line 13: is there something missing from this  comment?

bryophyta (approved):
I'm not too familiar with the details of how vite/babel/emotion interact,  but from what  I can tell this looks fine to me unless you have any  specific concerns about it?

On this sort of topic -- is there any testing  that  we could look at adding which would allow us to be more confident in making changes to this  library, given that  we typically don't interact with it all  that much?

vite.config.ts
mxdvl:
```suggestion
		// When using the “classic” JSX runtime in the react plugin,
		// we need to set these options: https://esbuild.github.io/api/#jsx-dev
```

mxdvl:
> On this sort of topic -- is there any testing that we could look at adding which would allow us to be more confident in making changes to this library, given that we typically don't interact with it all that much?

Deploying this a a static site on Github pages would be relatively easy!

mxdvl:
@bryophyta this was branched off #633, which still needs a review before merging!

PR #633, opened 2022-10-28
Chores – clean up code

mxdvl:
## What does this change?

Remove `React` imports & format everything with prettier.

## Why?

The imports are implicit. Prettier is only helpful if enforced.


mxdvl:
Merging this should release v11.0.2. It should have been merged into `main` before #632, ideally, but I was too hasty in clicking the green button.

bryophyta (approved):
this looks good from what I can see!

PR #658, opened 2022-11-22
Correctly assign dev and peer dependencies

mxdvl:
## What does this change?

Pin our development dependencies to lowest of the specified peer dependencies. Ajust our scripts that checks this remains true.
## Why?

This is the recommended way to dealing with dependencies, as [outlined in our docs](https://github.com/guardian/recommendations/blob/main/npm-packages.md#peerdependencies).

No changeset is needed as we have not touched our peer dependencies. This would however ensure that we don’t expect higher versions that specified.


bryophyta (commented):
The code looks good here, although I'm not sure I understand why the changes to the `yarn.lock` file are here?

mxdvl:
> The code looks good to me, although I'm not sure I understand why the changes to the `yarn.lock` file are here?

Good point. I assumed that’s because the specified ranges had changed, but it seems to have been some `yarn` weirdness.

bryophyta:
> > The code looks good to me, although I'm not sure I understand why the changes to the `yarn.lock` file are here?
> 
> The `yarn.lock` files need to be updated, because we’re no longer pointing to a range for our development dependencies. `react` is exactly 17.0.1. This is the correct way to develop locally, because we need to make sure that the project build with the lowest possible version of React 17 to tell consumers that they can have any compatible version.

Got it, thanks!

bryophyta (approved):

Issue #677, opened 2023-01-11
Soft-code identity url in discussion-rendering according to environment


ioannakok:
1. A user who does not have a discussion username set
2. Tries to make a comment
3. They're taken to the 'set username' flow
4. When they hit submit, nothing happens on the page, and the POST call to https://idapi.theguardian.com/user/me?api-key=dotcom-rendering fails with a CORS error

![Image](https://user-images.githubusercontent.com/19683595/213485129-2f8415a9-ca8d-4c42-bc20-1c99439878be.png)

5. The payload of that call is the setting of the new username, which means the call to set a new username doesn't work

![Image](https://user-images.githubusercontent.com/19683595/213485215-0b1527d8-1394-408f-8b38-26250048a3c1.png)

VDuczekW:
George and Ioanna are pairing and in flight

PR #679, opened 2023-01-13
Soft-code the identity url

ioannakok:
Co-authored-by: Georges Lebreton <georges.lebreton@guardian.co.uk>

## Context
In order to support the identity's team work on migration we need to make sure they can use the CODE environment for testing. Currently, the flow below is broken:

1. A user who does not have a discussion username set
2. Tries to make a comment
3. They're taken to the 'set username' flow
4. When they hit submit, nothing happens on the page, and the POST call to `https://idapi.theguardian.com/user/me?api-key=dotcom-rendering` fails with a CORS error

image

image

5. The payload of that call is the setting of the new username, which means the call to set a new username doesn't work 

image

## What does this change?
It decides the identity url based on the stage.

## Why?
Until now the https://idapi.theguardian.com/user/me was hard-coded and was being hit regardless of environment.
For example, currently COSE (https://m.code.dev-theguardian.com/) is hitting the PROD identity API endpoint but it is not one of the allowed domains listed [in the PROD configuration](https://github.com/guardian/identity/blob/main/identity-api/src/main/resources/PROD.conf#L11). [CODE identity api configuration](https://github.com/guardian/identity/blob/main/identity-api/src/main/resources/CODE.conf#L11) includes https://m.code.dev-theguardian.com/ in the valid guardian domains.

## Link to supporting GH issue
#677  https://github.com/guardian/dotcom-rendering/issues/6665

### DIFF
--- a/.changeset/hungry-pears-behave.md
+++ b/.changeset/hungry-pears-behave.md
@@ -1,0 +1,5 @@
+---
+'@guardian/discussion-rendering': patch
+---
+
+Soft-code identity url in set username flow
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -45,6 +45,7 @@ type Props = {
 	) => Promise<CommentResponse>;
 	onPreview?: (body: string) => Promise<string>;
 	onExpand?: () => void;
+	stage?: string;
 };
 
 const footerStyles = css`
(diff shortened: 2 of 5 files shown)


jamesgorrie (commented):
Where is the stage value actually coming from?

ioannakok:
Closing this. We'll create another PR

ioannakok:
This PR got superseded by: https://github.com/guardian/discussion-rendering/pull/680

PR #680, opened 2023-01-16
Discussion-rendering consumes `idApiUrl`

ioannakok:
Co-authored-by: Georges Lebreton <georges.lebreton@guardian.co.uk>

## Context
In order to support the identity's team work on migration we need to make sure they can use the CODE environment for testing. Currently, the flow below is broken:

1. A user who does not have a discussion username set
2. Tries to make a comment
3. They're taken to the 'set username' flow
4. When they hit submit, nothing happens on the page, and the POST call to `https://idapi.theguardian.com/user/me?api-key=dotcom-rendering` fails with a CORS error

image

image

5. The payload of that call is the setting of the new username, which means the call to set a new username doesn't work 

image

## What does this change?
Adds `idApiUrl` prop to discussion-rendering `App`. Its value will be coming from CAPI and passed to discussion-rendering [via DCR](https://github.com/guardian/dotcom-rendering/pull/6968).

## Why?
CAPI knows the correct url

| CODE | PROD |
|------|------|
| image | image |

## Link to supporting GH issue
#677

### DIFF
--- a/src/lib/api.tsx
+++ b/src/lib/api.tsx
@@ -17,6 +17,7 @@ let options = {
 	baseUrl: 'https://discussion.theguardian.com/discussion-api',
 	apiKey: 'discussion-rendering',
 	headers: {},
+	idApiUrl: 'https://idapi.theguardian.com',
 };
 
 let defaultParams = {
@@ -27,14 +28,17 @@ export const initialiseApi = ({
 	baseUrl,
 	additionalHeaders,
 	apiKey,
+	idApiUrl,
 }: {
 	baseUrl: string;
 	additionalHeaders: AdditionalHeadersType;
 	apiKey: string;
+	idApiUrl: string;
 }) => {
 	options.baseUrl = baseUrl || options.baseUrl;
 	options.headers = additionalHeaders || options.headers;
 	options.apiKey = apiKey || options.apiKey;
+	options.idApiUrl = idApiUrl || options.idApiUrl;
 
 	defaultParams['api-key'] = options.apiKey;
 };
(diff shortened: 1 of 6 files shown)


.changeset/strong-eyes-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/discussion-rendering': patch
```
bryophyta, line 2: if we're adding a new required property to the app, should this be a major change instead?

src/lib/api.tsx
```diff
@@ -27,14 +28,17 @@ export const initialiseApi = ({
...
 }: {
 	baseUrl: string;
 	additionalHeaders: AdditionalHeadersType;
 	apiKey: string;
+	idApiUrl: string;
 }) => {
 	options.baseUrl = baseUrl || options.baseUrl;
 	options.headers = additionalHeaders || options.headers;
 	options.apiKey = apiKey || options.apiKey;
+	options.idApiUrl = idApiUrl || options.idApiUrl;
```
bryophyta, line 41: is this fallback intended to guard against the case where the `idApiUrl` passed on line 31 is an empty string?

bryophyta (commented):
this looks good to me overall! just a question about the semantic level of the change I think, and also I thought it would be good to clarify the intended use for the fallback value in `api.tsx`

.changeset/strong-eyes-explode.md
ioannakok: Yes, good point! I have updated the changeset.

src/lib/api.tsx
ioannakok: Good question! Yes, I believe that would be the empty string. I am not entirely clear about the full flow, i.e. how this value ends up in CAPI, but I'm guessing an empty string could happen, so we're adding this default case. Also, this was we are following the existing pattern.

.changeset/strong-eyes-explode.md
bryophyta: thanks!

bryophyta (approved):
looks good to me, thanks for clarifying the earlier questions!

PR #683, opened 2023-02-01
feat: enforce healthy packages

mxdvl:
## What does this change?

**BREAKING CHANGE** `@guardian/libs` peer dependency is now v12

Feature: add the health package checker from https://github.com/guardian/actions-npm-dependencies/

## Why?

Help reduce peer dependencies issues with consumers.


joecowton1:
Looks good when your checks are passing :)

joecowton1 (approved):

mxdvl:
This has introduced two lockfiles by mistake, thanks for flagging @OllysCoding.

The `pnpm-lock.yaml` should not have been introduced, but at least one of them should be removed.

PR #697, opened 2023-04-24
Vite 4 & Vitest

mxdvl:
## What does this change?

Use [Vitest](https://vitest.dev/) instead of Jest

## Why?

Faster tests, better integrated with Vite.


mxdvl:
I did check that the build output was identical locally:

```sh
$ md5 build/App.esm.js 
MD5 (build/App.esm.js) = bc4af9eed177b1ba291632d26ea00a2d
```

PR #710, opened 2023-05-12
Pin dependencies to exact versions & enforce it

mxdvl:
## What does this change?

It’s several changes related to `package.json` health, if people prefer individual PR, happy to oblige.

- pin dependencies to exact versions
- run test without watch
- bump `@types/react` & `@types/react-dom` versions
- update healthy package.json check

## Why?

The current recommendation is to use exact versions.

The current test command prevents the prepush check from terminating.

New compatible version of `@types/react` and `@types/react-dom` have been published.

The latest version of the package health script has a simpler syntax for known issues.


mxdvl:
Hard to keep this up to date, but it’s still the right thing to do at some point…

PR #726, opened 2023-06-21
Update TS to v5.13 along with Guardian packages

joecowton1:
## What does this change?

Upgrades to TS latest version, as well as updating all Guardian packages

## Why?

Dept wide package updates for new TS version.


OllysCoding (approved):
+1 LGTM

